PR #12, opened 2021-06-07
Bump rush pnpm

aruniverse:

PR #13, opened 2021-06-14
bump repo to use cra 4.0, ts ~4.3

aruniverse:
Bump to bentley/react-scripts 4.0
Had to update linting configs a bit, but they should still work like before


packages/modules/cra-template-web-viewer/template.json
```diff
@@ -30,12 +30,8 @@
       "history": "^4.10.1",
       "react": "16.13.1",
       "react-dom": "16.13.1",
-      "@svgr/webpack": "4.3.3",
-      "@testing-library/jest-dom": "^4.2.4",
-      "@testing-library/react": "^10.4.4",
-      "@testing-library/user-event": "^7.1.2",
+      "@svgr/webpack": "^5.5.0",
```
aruniverse, line 33: @kckst8 I don't think we need to manually depend on this ourselves. This is a direct-dep in cra, but there seems to be an issue where this pkg doesnt get installed with rush, so we need to explicitly depend on it in a rush repo, but for the everyday user this is unnecessary

packages/apps/web-viewer-test/package.json
```diff
@@ -38,18 +37,15 @@
...
+    "@svgr/webpack": "^5.5.0",
     "@types/history": "^4.7.6",
-    "@types/jest": "^26.0.4",
     "@types/node": "^14.0.19",
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
     "@types/react-router": "^5.1.8",
     "@types/react-router-dom": "^5.1.5",
-    "typescript": "^3.7.2"
+    "typescript": "~4.1.0"
```
aruniverse, line 48: i didnt bump the ts ver across the repo, i could if you like, but this is the ver core is now using and thought itd be help to stay on the same ver

aruniverse:
fyi, 4.0.3-dev.1 works as expected here, but will break once #12 goes through, due to an issue with the new CopyStaticAssetsPlugin having trouble traversing pnpm workspaces

aruniverse:
> fyi, 4.0.3-dev.1 works as expected here, but will break once #12 goes through, due to an issue with the new CopyStaticAssetsPlugin having trouble traversing pnpm workspaces

edit: we had found a workaround for this, but then found an issue with pnpm 6.x and duplicated peer deps, so this comment can be ignored

packages/apps/desktop-viewer-test/src/backend/main.ts
```diff
@@ -29,6 +29,10 @@ const viewerMain = async () => {
   Logger.setLevel(AppLoggerCategory.Backend, LogLevel.Info);
 
   const clientId = getAppEnvVar("CLIENT_ID") ?? "";
+  assert(
```
aruniverse, line 32: @calebmshafer @Carl-Hinkle @kckst8 

Any objections?

calebmshafer: I haven't looked through the rest of the template but wouldn't this be caught the same way we catch client ids in web now, via the environment vars?

aruniverse: afaik, we don't actually check what kind of client is being used in web either. 
we noticed the other day that you could theoretically use a spa client id with the desktop viewer atm, and wanted to prevent that and made sure ppl were using the right client type for their app.

kckst8: whats the harm in using a spa-**** client? Once the desktop viewer has proper local briefcase support, I assume native-xxx will be required, but for now its just snapshots or online.

packages/apps/web-viewer-test/package.json
```diff
@@ -63,5 +59,10 @@
     "start": "react-scripts --max_old_space_size=8192 start",
     "build": "",
     "test": ""
+  },
+  "eslintConfig": {
```
kckst8, line 63: why do we need this now? Ideally all eslint config would be in the scripts package (where the pre-commit script lives)

packages/modules/cra-template-desktop-viewer/template/.env
```diff
@@ -3,11 +3,15 @@ ITWIN_VIEWER_SCOPE ="email rbac-user:external-client projectwise-share product-s
 ITWIN_VIEWER_CLIENT_ID=""
 ITWIN_VIEWER_REDIRECT_URI=""
 ITWIN_VIEWER_SNAPSHOT="./samples/house_model.bim"
-DISABLE_ESLINT=true
+
+# Advanced CRA Config: https://create-react-app.dev/docs/advanced-configuration/
```
kckst8, line 7: great idea to add these links

packages/apps/desktop-viewer-test/src/backend/main.ts
aruniverse: No harm atm, just wanted to be pro-active about this. As long as later on when we do require a desktop client id, the user knows they need one, and not one for a spa

packages/apps/web-viewer-test/package.json
aruniverse: So theoretically we don't even need this right now, because we set `DISABLE_ESLINT_PLUGIN` in our .env file. 
But if its enabled, by default the react-app uses the default eslint rules, and it was complaining about usage of static variables and a few others, and i couldn't find the actual rule we were violating. But after setting the eslint config to point to our global repo rules, it seemed to work. 

So left this in just in case someone wanted to enable the eslint plugin in cra.

packages/apps/desktop-viewer-test/src/backend/main.ts
kckst8: @calebmshafer any thoughts on this? Guess I'm not sure what the difference is outside of knowing that they registered a desktop app for internal reporting. Guess if that is the case we should enforce it to have an accurate representation.

packages/apps/web-viewer-test/package.json
kckst8: Ok. If that is the case though, shouldn't it be added to template.json instead? The test apps won't affect the CRA template

packages/apps/desktop-viewer-test/src/backend/main.ts
calebmshafer: There is an explicit different between the two and I'm fairly certain using a spa client wouldn't work since technically that uses Auth Code flow and the native clients are registered as Auth Code + PKCE support.

That being said, I do not think we really need the assert as we don't valid that the web viewer is a spa so I see no reason to do the same for Desktop.  If our docs do not lead you down the correct path, then those need to be fixed rather than using this as a blocker..

packages/apps/web-viewer-test/package.json
aruniverse: yea this was a concern, and something i wanted to test with the new template after this went through, but ill update accordingly now

aruniverse: verified, this is not a concern by defautl if the app's eslintrc extends react-app, will update accordingly files

Carl-Hinkle (approved):

calebmshafer (approved):

PR #18, opened 2021-06-30
desktop cra 1.0

kckst8:
- Fix browserslist. 8.0.0 is the highest electron goes (its what core uses)
- upgrade desktop viewer to 1.0.1
- Remove prerelease text


aruniverse (approved):
As long as we're ok moving to 1.0 with the desktop template

PR #19, opened 2021-07-06
update web viewer min version in CRA template

kckst8:


aruniverse (approved):
youll need rush change

PR #20, opened 2021-07-07
fix no viewstate error

kckst8:
Use core ViewCreator that creates a default view instead of throwing an error if one does not exist in the model


packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -184,35 +181,17 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
...
-
           // attempt to construct a default viewState
-          const savedViewState = await ViewCreator.createDefaultView(
-            imodelConnection,
-            undefined,
-            viewIds.length > 0 ? viewIds[0] : undefined
-          );
+          const viewCreator = new ViewCreator3d(imodelConnection);
+
+          const options: ViewCreator3dOptions = viewCreatorOptions
```
aruniverse, line 187: Since you set useSeedView in the conditional below couldn't you just do:

```ts
const options: ViewCreator3dOptions = viewCreatorOptions ?? {};
```

aruniverse (approved):
lgtm code wise, just need to test it out, but trust you to merge it in if you think its ready as is

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
kckst8: could sure, but why not initialize the value when we know it will be true in that case? the setter below was for the condition where the user passes options, but no useSeedView. Also didn't want to mutate a prop, thus the spread operator

kckst8:
> lgtm code wise, just need to test it out, but trust you to merge it in if you think its ready as is

if you get time to test it, that would be great. I tested with with models that previously errored and models that worked. Another set of eyes would be helpful (just ensure views load as expected for a few models) since this is potentially catastrophic if it didn't work :-)

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
```diff
@@ -0,0 +1,333 @@
+/*---------------------------------------------------------------------------------------------
```
aruniverse, line 1: Nit: double header

I know this will get replaced/removed when core is updated so probably not worth fixing

aruniverse (approved):
LGTM. tested with the models with older schemas that were broken

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
kckst8: fixed

PR #23, opened 2021-07-15
Fix default view configuration

kckst8:
1) Allow for a custom function to be executed on the viewPort
2) Failing that, if the model has a default view configured, honor it
3) Failing that, execute the fit to view tool

NOTES:
1) This will diverge our view creator from the core version, but I believe it is needed. I will investigate with the core owners if we should port these changes there.
2) If/when this is merged and published, I will update the CRA templates to use a custom viewportConfigurer function that executes the fitview tool with a comment that it can be removed to honor default views. I think fitview provides the best view of all of the models that I tested, so it should still be the default, but this will allow consumers to easily delete a few lines if they want to honor default views instead

### DIFF
--- a/packages/apps/web-viewer-test/src/components/home/AuthConfigHome.tsx
+++ b/packages/apps/web-viewer-test/src/components/home/AuthConfigHome.tsx
@@ -113,15 +113,10 @@ export const AuthConfigHome: React.FC = () => {
       });
     };
 
-    tileTreesLoaded()
-      .then(() => {
-        IModelApp.tools.run(FitViewTool.toolId, viewPort, true);
-        viewPort.view.setStandardRotation(StandardViewId.Iso);
-      })
-      .catch(() => {
-        IModelApp.tools.run(FitViewTool.toolId, viewPort, true);
-        viewPort.view.setStandardRotation(StandardViewId.Iso);
-      });
+    tileTreesLoaded().finally(() => {
+      IModelApp.tools.run(FitViewTool.toolId, viewPort, true);
+      viewPort.view.setStandardRotation(StandardViewId.Iso);
+    });
   };
 
   return (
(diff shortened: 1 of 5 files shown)


packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
```diff
@@ -86,17 +78,68 @@ export class ViewCreator3d {
...
+            // after 20 seconds, stop waiting and fit the view
+            if (now.getTime() - start.getTime() > 20000) {
+              reject();
+            }
+          }, 100);
+        });
+      };
+
+      tileTreesLoaded()
+        .then(() => {
```
aruniverse, line 130: Why not `.finally` if both the `.then` and `.catch` does the same thing?

packages/apps/web-viewer-test/src/components/home/AuthConfigHome.tsx
```diff
@@ -89,6 +94,36 @@ export const AuthConfigHome: React.FC = () => {
     }
   };
 
+  const viewConfiguration = (viewPort: ScreenViewport) => {
```
aruniverse, line 97: isn't this the default behavior if you don't provide `standardViewId`?

kckst8: which line? The whole function? It would not be the default if the model has a default view id.....This is basically saying to fit the view and set the standard rotation to Iso regardless

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
kckst8: yes definitely

packages/apps/web-viewer-test/src/components/home/AuthConfigHome.tsx
aruniverse: Ah yes thats right, sorry disregard this comment. I guess if anything you could also change this to be .finally as well

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
kckst8: updated

aruniverse (approved):
tested the updated query that looked good. code looks good too

PR #28, opened 2021-07-23
Update templates and test apps to support different auth issuers and the itwinjs scope

aruniverse:


packages/apps/desktop-viewer-test/.env
```diff
@@ -1,8 +1,9 @@
 # ---- Configuration ----
-ITWIN_VIEWER_SCOPE ="email rbac-user:external-client projectwise-share product-settings-service context-registry-service:read-only imodelhub organization profile openid"
+ITWIN_VIEWER_SCOPE ="email organization profile openid itwinjs"
 ITWIN_VIEWER_CLIENT_ID=""
 ITWIN_VIEWER_REDIRECT_URI=""
 ITWIN_VIEWER_SNAPSHOT="./samples/house_model.bim"
+ITWIN_VIEWER_ISSUER_URL=""
```
calebmshafer, line 6: I'd set the default Issuer to be ims.bentley.com now since you're using that scope by default. Then just provide the option to review back...

calebmshafer (approved):

packages/apps/desktop-viewer-test/src/backend/main.ts
```diff
@@ -42,6 +43,7 @@ const viewerMain = async () => {
       scope,
       redirectUri:
         redirectUri && redirectUri.length > 0 ? redirectUri : undefined,
+      issuerUrl,
```
kckst8, line 46: what happens if this is empty/undefined?....I believe it will end up as an empty string as is and may not be handled as falsey in imjs, resulting in an error.  This was the case for redirectUri (overriding the default with empty string), thus the additional condition. I would check what happens in this case and maybe account for it as needed

packages/modules/cra-template-desktop-viewer/template/src/backend/main.ts
```diff
@@ -42,6 +43,7 @@ const viewerMain = async () => {
       scope,
       redirectUri:
         redirectUri && redirectUri.length > 0 ? redirectUri : undefined,
+      issuerUrl,
```
kckst8, line 46: see comment in test app

packages/apps/desktop-viewer-test/src/backend/main.ts
aruniverse: Great catch! Fixed

common/changes/@itwin/cra-template-desktop-viewer/feat-imsAuth_2021-08-04-14-06.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/cra-template-desktop-viewer",
+      "comment": "Default to using ims for issuer",
```
aruniverse, line 5: Will add a more detailed changelog entry

Carl-Hinkle (commented):
lgtm

kckst8 (approved):

PR #30, opened 2021-07-27
Not signed in

spap975:
Added a simple text box to the BaseViewer that displays "Not Signed In" if the user has not been authorized yet :)


packages/modules/viewer-react/src/components/BaseViewer.scss
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+@import "~@bentley/ui-core/lib/ui-core/style/themecolors.scss";
+
+.not-signed-in-container {
+    width: 200px;
```
Carl-Hinkle, line 8: Might be better to use flexbox than absolutely positioning this.  We never know what the size of the viewer's container is going to be.

Carl-Hinkle (changes requested):

packages/modules/viewer-react/src/components/BaseViewer.scss
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+@import "~@bentley/ui-core/lib/ui-core/style/themecolors.scss";
+
+.not-signed-in-container {
```
Carl-Hinkle, line 7: Might be better to use flexbox than absolutely positioning this. We never know what the size of the viewer's container is going to be.

packages/modules/viewer-react/src/components/BaseViewer.tsx
```diff
@@ -107,6 +114,11 @@ export const BaseViewer: React.FC<ViewerProps> = ({
   // TODO not signed in message / loader
   return (
     <ErrorBoundary>
+      {!authorized && (
+        <div className="not-signed-in-container">
```
aruniverse, line 118: Could you look at using the following component?
import { FillCentered } from "@bentley/ui-core";

Wonder how this will look and if its easier to use.

aruniverse (commented):
You're missing rush changes.

Could you add a screenshot of what your current not signed in container looks like?

spap975:
image

Carl-Hinkle:
Can you change that to "Please sign in."

With sentence case.

packages/modules/viewer-react/src/components/BaseViewer.tsx
```diff
@@ -107,6 +108,9 @@ export const BaseViewer: React.FC<ViewerProps> = ({
   // TODO not signed in message / loader
```
aruniverse, line 108: remove the comment

packages/modules/viewer-react/src/components/BaseViewer.tsx
```diff
@@ -107,6 +108,9 @@ export const BaseViewer: React.FC<ViewerProps> = ({
...
   return (
     <ErrorBoundary>
+      {(!authorized || !iModelJsInitialized) && (
```
aruniverse, line 111: if authorized and imodeljs not initialized, we show "please sign in" which makes no sense.

image

Wonder if we should only initialize imjs after youre authorized, and we should show a spinner when initializing

aruniverse: lets disregard delaying the initialization to after being authenticated, i dont like the coupling we have here with IModelApp.authorizationClient and imjsInitialization.

common/changes/@itwin/viewer-react/notSignedIn_2021-07-27-17-02.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/viewer-react",
+      "comment": "Added a component displaying Not Signed In when the user isn't authorized",
+      "type": "minor"
```
aruniverse, line 6:
```suggestion
      "type": "patch"
```

aruniverse:
Could you update the BaseBlankViewer as well? 
packages\modules\viewer-react\src\components\BaseBlankViewer.tsx

packages/modules/viewer-react/src/components/BaseBlankViewer.tsx
```diff
@@ -3,7 +3,13 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
aruniverse, line 7: duplicate header

aruniverse (approved):

PR #31, opened 2021-07-29
Change type of WebAuthorizationOptions.oidcClient to FrontendAuthorizationClient

aruniverse:
Change type of WebAuthorizationOptions.oidcClient to [FrontendAuthorizationClient](https://www.itwinjs.org/reference/frontend-authorization-client/authorization/frontendauthorizationclient/). 

A weaker type than the previously used [BrowserAuthorizationClient](https://www.itwinjs.org/reference/frontend-authorization-client/browserauthorization/browserauthorizationclient/) but this is the same type as [IModelApp.authorizationClient](https://www.itwinjs.org/reference/imodeljs-frontend/imodelapp/imodelappoptions/)

This change is being done to accommodate the proxy-viewer use case, but this should also make it much easier for users to switch from the deprecated bentley/itwin-viewer-react package to this one.

A larger discussion needs to be had regarding auth configurations for the viewer; because technically a valid config a user can pass into the viewer is an empty object, but that won't do much and you'll see the newly added "please sign in". For a case like this where you're proxying requests for the visualization apis like gpb then you really shouldn't need an auth client at all on the frontend wrt to the viewer.

Also export types from web-viewer-react pkg.

### DIFF
--- a/common/changes/@itwin/web-viewer-react/fix-proxy_2021-07-29-23-05.json
+++ b/common/changes/@itwin/web-viewer-react/fix-proxy_2021-07-29-23-05.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/web-viewer-react",
+      "comment": "change type of WebAuthorizationOptions.oidcClient to FrontendAuthorizationClient",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@itwin/web-viewer-react",
+  "email": "aruniverse@users.noreply.github.com"
+}
--- a/packages/modules/web-viewer-react/src/services/ItwinViewer.ts
+++ b/packages/modules/web-viewer-react/src/services/ItwinViewer.ts
@@ -17,7 +17,7 @@ import ReactDOM from "react-dom";
 import { ViewerFrontstage } from "..";
 import { Viewer } from "../components/Viewer";
 import {
-  ItwinViewerParams,
+  ItwinWebViewerParams,
   WebAuthorizationOptions,
   WebViewerProps,
 } from "../types";
@@ -42,7 +42,7 @@ export class ItwinViewer {
 
   onIModelConnected: ((iModel: CheckpointConnection) => void) | undefined;
 
-  constructor(options: ItwinViewerParams) {
+  constructor(options: ItwinWebViewerParams) {
     if (!options.elementId) {
       //TODO localize
       throw new Error("Please supply a root elementId as the first parameter"); //TODO localize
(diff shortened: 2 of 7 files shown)


common/changes/@itwin/web-viewer-react/fix-proxy_2021-07-29-23-05.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/web-viewer-react",
+      "comment": "change type of WebAuthorizationOptions.oidcClient to FrontendAuthorizationClient",
+      "type": "minor"
```
aruniverse, line 6: Couldnt decide between minor/patch. Feel free to disagree with what i have here and i'll change it.

Carl-Hinkle:
Any idea when FrontendAuthorizationClient will be made public?

common/changes/@itwin/web-viewer-react/fix-proxy_2021-07-29-23-05.json
Carl-Hinkle: I think minor makes sense.

aruniverse:
> Any idea when FrontendAuthorizationClient will be made public?

I believe the Auth workflow and clients will be refactored in 3.0 so probably not for a while. Caleb or Raman might know more.

Carl-Hinkle (commented):
Since you're changing exported interface names, this should probably be a major version bump.  The changelog should also describe the migration process.

packages/modules/web-viewer-react/src/services/ItwinViewer.ts
```diff
@@ -17,7 +17,7 @@ import ReactDOM from "react-dom";
 import { ViewerFrontstage } from "..";
 import { Viewer } from "../components/Viewer";
 import {
-  ItwinViewerParams,
```
kckst8, line 20: I believe we can just delete iTwinViewerParams interface from the base viewer-react package. It is web-specific now. Looks like it was just left behind once I moved it to the web package. 

Once we do that, we can leave the name the same (iTwinViewerParams) but just export it from the web package now. It will no longer be a breaking change in the web and the breaking change in the base is unlikely to be used in desktop apps so we should be good there as well.

kckst8 (changes requested):

packages/modules/web-viewer-react/src/services/ItwinViewer.ts
aruniverse: So are we doing a major ver bump of the base-viewer pkg? And that will propagate to the web and desktop viewer pkgs as minor bump?

kckst8: I don't think we need a major bump anywhere. We don't advocate using the base package anywhere, and even if somebody did, they would not be using that interface. The only potential breakage is that this interface would no longer be exposed to the desktop. However, its only use case is for the js (CDN) web use case, so no issue there

kckst8 (approved):

PR #32, opened 2021-07-30
fixed the height of the viewer when app insights are on

spap975:
Previously, when including a value for app insights in the .env file, the height of the viewer would shrink.
image
To fix this, I simply added a class name to the div that was giving us trouble and styled it to have a height of 100%.
image


aruniverse (changes requested):
rush changes, patch

aruniverse (approved):

PR #35, opened 2021-08-05
proxy ipc calls to make it easier to add new ones

MichaelBelousov:
I found it easier to add new calls ipc calls with this proxy, and figured I would share to see if it makes sense to upstream.
My application requires adding ipc calls such as, for instance, opening a save file dialog, and a few others, this change made it more manageable to add several.

I also found that using the `no-unused-vars` linter rule causes a false positive on arrow function type notation in typescript, so I turned it off and instead enabled the @typescript-eslint version of it.

### DIFF
--- a/packages/modules/cra-template-desktop-viewer/template/src/frontend/app/ITwinViewerApp.ts
+++ b/packages/modules/cra-template-desktop-viewer/template/src/frontend/app/ITwinViewerApp.ts
@@ -33,6 +33,9 @@ export class ITwinViewerApp {
     return IModelApp.i18n.translate(`iTwinViewer:${key}`, options);
   }
 
+  // This proxy object forwards any method calls to the backend over IPC.
+  // This way, you can call all ipc methods like `ITwinViewerApp.ipcCall.openFile(args)`
+  // Any new backend/ipc methods you need should be added to the ViewerIpc interface, and then implemented in the ViewerHandler
   public static ipcCall = new Proxy({} as IpcMethods, {
     get(_target, key: keyof IpcMethods): AsyncFunction {
       const makeIpcCall =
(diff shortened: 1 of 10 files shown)


aruniverse:
Got this:
image
Then after i added the following in tsconfig.backend.json 
```
    "esModuleInterop": true,
```

i got this error:
image

MichaelBelousov:
my bad, I did change minimist in my app, I haven't checked why it worked before.
I also didn't get any errors for that getConfig call when `rush build`ing, but I see it in my editor, so I'll fix and run the app

aruniverse (approved):
LGTM, but would like approval from @kckst8 before this gets merged in.

This definitely helps simplify things, write less code, but wonder if the proxy is going to confuse users.

kckst8 (commented):
Thanks @MichaelBelousov ! Agree with @aruniverse that this is a helpful change. But also agree that it is not intuitive for new users. Could you please add some function comments to the files in the cra template to help users understand what the function does and how to add ipc methods? No need in the test app.

packages/modules/cra-template-desktop-viewer/template/src/frontend/app/ITwinViewerApp.ts
```diff
@@ -16,27 +16,47 @@ import {
...
-    ) as PromiseReturnType<ViewerIpc[T]>;
-  }
-  public static async getConfig(): Promise<ViewerConfig> {
-    if (!this.config) {
-      this.config = await this.callBackend("getConfig");
-    }
-    return this.config;
-  }
+
+  // This proxy object forwards any method calls to the backend over IPC.
```
MichaelBelousov, line 36: @kckst8 is this good?

MichaelBelousov: there are 3 lines of comments, they don't seem to appear in the context

kckst8 (approved):

kckst8 (commented):
lgtm but needs rush change for the cra template pkg before merge

PR #40, opened 2021-08-17
Initializer cancel

StefanRetief:
Fixed issue where initialization cancelling would not exit early and the viewer would continue initializing after the promise has already been rejected. Using generators, cancelling can exit early and after the last promise called has completed, resulting in fewer cases where the initializers between mounted and unmount viewers will clash.


packages/modules/desktop-viewer-react/src/utilities/MakeCancellable.ts
```diff
@@ -0,0 +1,57 @@
+/*---------------------------------------------------------------------------------------------
```
StefanRetief, line 1: Please advise on if this is the correct place for this file. Currently each package has their own where necessary, meaning lots of repeating code. However, I didn't see it fit to export this from a single location, unless it should be exported from the base package.

aruniverse: if its the same exact file in all 3 pkgs, i would add and export it in the base (viewer-react) and import accordingly in the web and desktop

packages/modules/viewer-react/src/utilities/MakeCancellable.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+        next(result);
+        return null;
+      }
+    }
+
+    function onRejected(err: any) {
+      let result;
+      try {
+        result = gen.throw(err);
```
kckst8, line 38: is this necessary? can't just reject and move on?

packages/modules/viewer-react/src/utilities/MakeCancellable.ts
```diff
@@ -0,0 +1,57 @@
...
+      reject({ reason: "cancelled" });
+    };
+
+    onFulfilled();
+
+    function onFulfilled(res?: any) {
+      if (!cancelled) {
+        let result;
+        try {
+          result = gen.next(res);
```
kckst8, line 26: possibly misunderstanding but is this attempting to pass the response from the previous iteration into the next iterator? Is that right?

StefanRetief: yes, similar to how ```then``` methods work on promises. This whole thing can be thought of a giant promise chain with checking between statements

packages/modules/viewer-react/src/utilities/MakeCancellable.ts
StefanRetief: rejecting the promise doesn't necessarily stop the generator from further executing if it is accessed again. The reject will only reject the current promise we're wrapped in, but doesn't prevent the generator function from sticking around in a suspended state, which could also cause a memory leak

kckst8 (approved):

PR #41, opened 2021-08-18
Fix viewstate race condition

kckst8:


packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -94,25 +96,69 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
...
+    }, [
+      connection,
+      viewportOptions,
+      blankConnectionViewState,
+      viewCreatorOptions,
+    ]);
+
+    useEffect(() => {
+      void getViewState();
+    }, [connection, viewportOptions, blankConnectionViewState, getViewState]);
```
aruniverse, line 160: missing viewCreatorOptions?

kckst8: yep. fixed

aruniverse (approved):
lgtm, approving, but would like final approval from @StefanRetief before merging

StefanRetief (approved):
Tested and works

PR #44, opened 2021-09-07
Fixed issues related to viewstate and connections

StefanRetief:
This fixes the issues related to Viewstate and connection changing.

- Changing imodel connections (contextid, imodelid, snapshot) will result in the viewstate being cleared and a default viewstate being fetched and applied.
- Rendering a view will only occur when both a viewstate and a connection has been established or a frontstage that does not require a connection has been provided.
- Viewstates created for the blank viewer will be more unopinionated, resulting in always giving a top-down view. Invalid ```lookAt``` properties will result in a console error. (Please advise).
- (Regression) Connections will be closed each time connections change.

It appears that there is currently a race condition with setting a frontstage provider. It appears that when the ```IModelViewer``` is loading, the ```FrontstageManager.setActiveFrontstage``` method cannot be called, as the desired frontstage will not be applied. Waiting for the the component to finish rendering its children (async driven), and then calling the ```setActiveFrontstage``` method will give the desired result.


packages/modules/viewer-react/src/components/iModel/IModelViewer.tsx
```diff
@@ -40,7 +40,10 @@ export const IModelViewer: React.FC<ModelProps> = ({
     // set the active frontstage to the current default
     if (defaultFrontstage) {
       // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      FrontstageManager.setActiveFrontstageDef(defaultFrontstage.frontstageDef);
+      setImmediate(
```
kckst8, line 43: unfortunately, we cant use this. Lack of browser support and no plans to standardize:

https://caniuse.com/setimmediate

https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

packages/modules/viewer-react/src/tests/components/iModel/IModelViewer.test.tsx
```diff
@@ -18,6 +18,8 @@ jest.mock("@bentley/ui-framework");
 jest.mock("@bentley/ui-abstract");
 jest.mock("@microsoft/applicationinsights-react-js");
 
+const flushPromises = () => new Promise(setImmediate);
```
kckst8, line 21: same here

kckst8 (commented):
Added a comment. Other code looks ok but I will test it today. 

I'm in mid refactor and making some changes in this area as well. I think these changes will patch some larger issues in the IModelLoader component. There are too many asynchronous tasks that are related with potential collisions and/or improper state management.

kckst8 (changes requested):
Changing the iModel id causes an error when closing the connection:

image

You can reproduce using the "Switch iModel" button in the auth config route in the web viewer test app

kckst8 (commented):
seems like a lot of file changes. A lot of scss files have duplicate copyright headers. Assuming from a merge commit? Possibly line ending/encoding?

StefanRetief:
> seems like a lot of file changes. A lot of scss files have duplicate copyright headers. Assuming from a merge commit? Possibly line ending/encoding?

Yeah, probably from the latest merge with master. I thought I caught all of them, but I'll resolve the last few.

kckst8 (approved):

PR #45, opened 2021-09-08
Desktop CRA / Viewer initialization refactoring

kckst8:
This is the beginning of the desktop template (and eventual app) refactor. I made the changes in the desktop-viewer-test app for validation. I still need to fix tests, docs, etc. My goal with this initial PR is to have few people review the code and pull and run the app locally to help identify issues early on. Ideally somebody could test on Windows since I am developing on a Mac (I intend to test in a VM as well). If all goes well, I will replace the current desktop template bits with these changes

To Test:

rush build 
followed by npm start in packages/apps/desktop-viewer-test

This is the first phase in order to limit major changes. 

Still TODO:

- recent files list (persistence and retrieval)
- download local briefcase via tile context menus (view remote menu option will be replaced with "download briefcase")
- support local briefcases in the tiles (show an indicator and use local when available), recent list, etc.
- theme switcher
- disable functions based on connectivity status
- add user preference to show most file recent on startup (in lieu of the start page)

Major changes as far as the base packages are concerned are mostly around initialization:

- added initialization hooks that can be used to initialize ITJS prior to rendering a viewer
- always terminate and re-initialize the statics per instance (with the exception of IModelApp and its subclasses). Due to the way that IModelHost interacts with NativeApp, it's not feasible to "restart" the app without restarting the host at this time. My expectation is that this may be possible in 3.0 depending on if we can de-couple a few more things. 

@MichaelBelousov, added you primarily to review the new hooks for your use case (see App.tsx for usage)

@StefanRetief added for review of new initialization/termination logic to ensure it does not break sandboxes


MichaelBelousov:
the new `useDesktopViewerInitializer` hook looks good for my use case, tomorrow I'll try to test using it in my app in case I run into anything

packages/apps/desktop-viewer-test/src/backend/main.ts
```diff
@@ -60,10 +63,81 @@ const viewerMain = async () => {
...
+  }
+
+  const menu = Menu.buildFromTemplate(template as MenuItemConstructorOptions[]);
+
+  Menu.setApplicationMenu(menu);
 };
 
 try {
-  viewerMain(); // eslint-disable-line @typescript-eslint/no-floating-promises
+  void viewerMain();
```
calebmshafer, line 140: I had the same comment on a PR that Keith recently made... I'm not a huge fan of this fix that ESLint recommends for this, it is an easy thing to miss when reviewing code and/or PRs of a `void` vs. and `await`. Plus not very obvious on what it actually does...

kckst8: I think this is the recommended approach to say this function is truly asynchronous and we do not need to wait on it. This is not a case where we can or want to use await or top level await. The only alternatives that I am aware of are empty catches or disabling the rule. I feel like void is more meaningful and cleaner than both of those? https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md

MichaelBelousov:
I'm not going to test it in my app for now because of time constraints on my end, I assume someone else will test it on windows. Please let me know if you need me to test something specific and I'll get to it when I can. The hooks looks appropriate for my use case

packages/apps/desktop-viewer-test/src/frontend/components/routes/HomeRoute.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { RouteComponentProps, useNavigate } from "@reach/router";
+import React from "react";
+
+import { useMenuHandlers } from "../../hooks";
+import Home from "../home/Home";
+
+//eslint-disable-next-line no-empty-pattern
+export const HomeRoute = ({}: RouteComponentProps) => {
+  const navigate = useNavigate();
```
aruniverse, line 14: should this be higher up? in App.tsx?

aruniverse: specifically the useMenuHandlers hook

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
```diff
@@ -0,0 +1,36 @@
...
+  ITwinsRoute,
+  SnapshotRoute,
+} from "./routes";
+
+const App = () => {
+  const initialized = useDesktopViewerInitializer();
+
+  return initialized ? (
+    <div style={{ height: "100%" }}>
+      <Router style={{ height: "calc(100% - 20px)" }}>
```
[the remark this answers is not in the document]
aruniverse: nvm, looks like thats needed, but cant you set the height to 100%?

packages/apps/desktop-viewer-test/src/frontend/components/signin/SignIn.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const SignIn = () => {
+  const [signingIn, setSigningIn] = useState(false);
+
+  const onSignInClick = () => {
+    setSigningIn(true);
+    IModelApp.authorizationClient?.signIn();
+  };
+
+  return (
+    <div className="signin-container">
```
aruniverse, line 21: instead of this signin container, you could use <FillCentered /> from ui-core

packages/apps/desktop-viewer-test/src/frontend/components/signin/SignIn.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <div className="signin">
+        <span className="icon icon-user signin-user" />
```
aruniverse, line 23: nbd, but is there an icon from itwinui we can use? noit sure how many people will realize these icons come from the webfont

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <HomeRoute path="/" />
+        <DownloadRoute path="itwins/:iTwinId/:iModelId" />
```
aruniverse, line 25: no leading `/` ?

packages/apps/desktop-viewer-test/src/frontend/components/viewer/ModelViewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+      });
+    };
+
+    tileTreesLoaded().finally(() => {
+      IModelApp.tools.run(FitViewTool.toolId, viewPort, true, false);
+      viewPort.view.setStandardRotation(StandardViewId.Iso);
+    });
+  }, []);
+
+  const onIModelAppInitialized = useMemo(() => {
```
aruniverse, line 51: does this really need to be memoized?

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.scss
```diff
@@ -0,0 +1,64 @@
...
+      margin: 0;
+    }
+  }
+
+  > .iui-tabs-wrapper {
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+    //position: absolute;
+    width: 100%;
```
aruniverse, line 43: could remove the width, theres a horizontal scrollbar added because of the margin 
image

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.tsx
```diff
@@ -0,0 +1,129 @@
...
+      ? PROJECT_TYPE_MAP.indexOf(window.location.search)
+      : 0
+  );
+
+  const [searchValue, setSearchValue] = React.useState("");
+  const [searchParam, setSearchParam] = React.useState("");
+  const startSearch = React.useCallback(() => {
+    setSearchParam(searchValue);
+  }, [searchValue]);
+  const [accessToken, setAccessToken] = useState<AccessToken | undefined>();
```
aruniverse, line 56: could just use the `useAccessToken` hook

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.tsx
```diff
@@ -0,0 +1,129 @@
...
+      const token = await IModelApp.authorizationClient?.getAccessToken();
+      setAccessToken(token);
+    };
+    void getAccessToken();
+  }, []);
+
+  return accessToken ? (
+    <>
+      <div className="itv-scrolling-container select-project">
+        <HorizontalTabs
```
aruniverse, line 70: The tabs and search aren't sticky, when i scroll they move off the window

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import "./SelectIModel.scss";
+
+import { IModelGrid, IModelGridProps } from "@itwin/imodel-browser-react";
+import { useNavigate } from "@reach/router";
+import React from "react";
+
+export const SelectIModel = ({ accessToken, projectId }: IModelGridProps) => {
```
aruniverse, line 12: when im on this pg theres no way for me to navigate back, probably cause the menu is missing on my end, but wondering if we need a header

aruniverse (dismissed):
Overall looks great, i need to investigate the missing windows menu bar, thats the only "blocker" atm

packages/modules/viewer-react/src/types.ts
```diff
@@ -105,6 +106,22 @@ export interface ItwinViewerInitializerParams {
   imodelClient?: IModelClient;
 }
 
+/**
+ * list of all properties needed for iTwin.js initialization
+ * This list should be updated each time a property is added/removed from ItwinViewerInitializerParams
+ */
+export const iTwinViewerInitializerParamList = [
```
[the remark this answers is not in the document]
kckst8: yeah agree but I don't know of anything to extract properties from a type/interface since they don't really exist at runtime

packages/modules/viewer-react/src/types.ts
```diff
@@ -83,6 +83,7 @@ export interface ItwinViewerCommonParams
     IModelLoaderParams {}
 
 export interface ItwinViewerInitializerParams {
+  [index: string]: any;
```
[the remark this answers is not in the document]
kckst8: so that we can use dynamic accessors (i.e. object[key]), in the getInitializationOptions utility function

packages/modules/viewer-react/src/utilities/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+  return true;
+};
+
+/**
+ * Pull out iTwin.js initialization options from a set of options
+ * @param options
+ * @returns
+ */
+export const getInitializationOptions = (options?: ViewerProps) => {
```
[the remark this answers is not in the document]
kckst8: yeah, but lodash is a bit heavy for just a single function. I'm not a huge fan of adding a large utility package unless we have multiple needs

packages/apps/desktop-viewer-test/src/frontend/components/viewer/ModelViewer.tsx
kckst8: I've found that all React callback/output props (i.e. functions) should be memoized or they can cause unnecessary renders

packages/modules/viewer-react/src/types.ts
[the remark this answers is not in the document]
MichaelBelousov: is it not possible to do `object[key]` without a typeless key signature as long as you cast `key` to `keyof typeof object`?

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
kckst8: the goal was to use menus for this (i.e. opening a new file/workspace in VS Code). It's controversial I know. I'd like to get others thoughts once I get the menus working on Windows. We could move to a sidebar, but I think its unnecessary real estate in a desktop app where menus are standard.

packages/modules/viewer-react/src/types.ts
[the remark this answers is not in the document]
kckst8: This could work. Thanks @MichaelBelousov. I guess my only argument is we still need to maintain a second list of the properties in the sample object.....at least they are type safe though

packages/modules/viewer-react/src/types.ts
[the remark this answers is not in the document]
kckst8: a cast would work, but it would need to be done in each instance of accessing props. Why not handle it in the type itself?

packages/modules/viewer-react/src/types.ts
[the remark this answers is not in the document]
MichaelBelousov: if there are some default params, you can store them as an object of the Params type in this scope (if you have defaults anywhere else before, such as in conditinoal logic of the function taking the param object, you can have those reference the `defaultOptions` object you place in this scope instead of inlining them). That may allow you to logically not maintain a second list, it will just have the sideeffect of also allowing you to generate a typesafe key list. Otherwise I agree it's better to have a duplicate list for the typesafe keys in this case.

packages/modules/viewer-react/src/types.ts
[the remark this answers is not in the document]
MichaelBelousov: if the object is never intended to have other properties other than those listed, then you're losing some typesafety, e.g. you can now mistype while writing `myObj["appInsightssssKey"]` and typescript won't tell you it won't have that and you may not notice depending on your tests. Relying on the `keyof Params` type, which you can pull out a new variable to assert the type or use it as the type of function parameters, will guarantee the typesafety, it's only sad that it's a tad more verbose than most typing constructs. I will read `getInitializationOptions` to be sure, but in general you shouldn't weaken the type system's understanding of an object's intended shape

packages/modules/viewer-react/src/utilities/index.ts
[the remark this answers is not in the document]
MichaelBelousov: aren't there per-function lodash packages just to prevent adding the whole of lodash while still getting the goodness?
https://www.npmjs.com/package/lodash.pick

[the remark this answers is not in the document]
kckst8: did not know this existed...guess people were tired of excess....thanks.

packages/modules/viewer-react/src/types.ts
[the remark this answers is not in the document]
kckst8: updated

packages/modules/viewer-react/src/hooks/useBaseViewerInitializer.tsx
```diff
@@ -0,0 +1,48 @@
...
+      void BaseInitializer.initialize(options).then(() => {
+        void BaseInitializer.initialized.then(() => {
+          setBaseViewerInitalized(true);
+        });
+      });
+    }
+    if (!isMounted.current) {
+      return BaseInitializer.cancel(); //TODO rename stopDesktopInitalizer
+    }
+  }, [options, delay, baseViewerInitOptions]);
```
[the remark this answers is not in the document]
kckst8: updated

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
kckst8: Headers should be working now on Windows. @calebmshafer This was due to ElectronHost attempting to force auto-hiding the menubar for some reason. I was able to workaround this.

Ultimately, if we are good with this UX, I think the real solution is a frameless window with a custom menu/title bar like VS Code (https://github.com/AlexTorresSk/custom-electron-titlebar). There is some effort there though that I think can wait until another minor version.

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.tsx
kckst8: done

packages/apps/desktop-viewer-test/src/backend/main.ts
```diff
@@ -60,10 +63,81 @@ const viewerMain = async () => {
   if (process.env.NODE_ENV === "development") {
     ElectronHost.mainWindow?.webContents.toggleDevTools();
   }
+  // add the menu
+  ElectronHost.mainWindow?.on("ready-to-show", createMenu);
+};
+
+const createMenu = () => {
```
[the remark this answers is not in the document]
kckst8: should be fixed

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
kckst8: by design.....so it isnt made a child of the current route (there are nested routes in reach router)

packages/apps/desktop-viewer-test/src/frontend/components/signin/SignIn.tsx
kckst8: prefer to not use ui-core, especially something this simple

packages/apps/desktop-viewer-test/src/frontend/components/signin/SignIn.tsx
kckst8: yep. switched to an itwinui icon

packages/apps/desktop-viewer-test/src/frontend/index.scss
```diff
@@ -3,20 +3,20 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 @import "~@bentley/ui-core/lib/ui-core/index";
+@import "@itwin/itwinui-css/scss/variables";
```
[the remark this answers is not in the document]
kckst8: i don't think we even need this import....famous last words

packages/modules/desktop-viewer-react/src/hooks/useConnectivity.tsx
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { InternetConnectivityStatus } from "@bentley/imodeljs-common";
+import { NativeApp } from "@bentley/imodeljs-frontend";
+import { useCallback, useEffect, useState } from "react";
+
+export const useConnectivity = () => {
```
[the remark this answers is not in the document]
kckst8: It will be. API wasn't working as I expected but I didn't want to remove it and start again from scratch

packages/apps/desktop-viewer-test/.env
```diff
@@ -1,8 +1,7 @@
 # ---- Configuration ----
-ITWIN_VIEWER_SCOPE ="email organization profile openid itwinjs"
+ITWIN_VIEWER_SCOPE ="imodels:read projects:read openid email profile organization itwinjs"
```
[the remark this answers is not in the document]
kckst8: is this documented somewhere on developer.bentley.com? We can add a link, but I'm not sure its the job of the template to maintain that documentation

packages/apps/desktop-viewer-test/package.json
```diff
@@ -65,23 +64,26 @@
...
     "@itwin/desktop-viewer-react": "^1.0.16",
-    "@reduxjs/toolkit": "^1.6.0",
     "electron": "^11.1.0",
     "minimist": "^1.2.5",
     "react": "16.13.1",
     "react-dom": "16.13.1",
-    "react-redux": "^7.2.0",
-    "dotenv-flow": "^3.2.0"
+    "dotenv-flow": "^3.2.0",
+    "@itwin/imodel-browser-react": "~0.11.0",
```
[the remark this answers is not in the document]
kckst8: done

packages/apps/desktop-viewer-test/src/frontend/components/routes/HomeRoute.tsx
kckst8: It needs the router functions so needs to be in a child of the router components. I thought about that as well, but the plan is to always start the app at the home route. Once we implement defaulting to most recent, that logic (the redirect) will go in the home route (at least in my head)

packages/modules/viewer-react/public/locales/en/iTwinViewer.json
```diff
@@ -15,5 +15,6 @@
   },
   "tools": {
     "walkTool": "Walk Tool"
-  }
+  },
+  "notInitialized": "IModelApp must be initialized prior to rendering the Base Viewer"
```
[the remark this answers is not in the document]
kckst8: lol. yep.....wish I could delete that line from everyone's memory. Guess that one will need to be de-localized.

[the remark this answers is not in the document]
kckst8: done

packages/apps/desktop-viewer-test/public/locales/en-PSEUDO/iTwinViewer.json
```diff
@@ -1,20 +1,15 @@
...
-  },
-  "backstage": {
-    "selectIModel": "Òpéñ Õñlíñë ìMõdêl",
-    "selectSnapshot": "Òpéñ Õfflìñê §ñãpshöt",
-    "mainFrontstage": "Vìéw ,çtìvé Mòdël"
-  },
-  "snapshotSelect": {
-    "title": "Òpéñ £öçàl §ñãpshöt",
-    "open": "$élêçt $ñàpshót Fìlé"
+  "downloadBriefcase": "Download Briefcase",
```
aruniverse, line 2: nbd, but not psuedo-localized, i think you just need to run the script again

packages/apps/desktop-viewer-test/public/links.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "textKey": "home.links.viewerComponent",
+    "url":"https://www.npmjs.com/package/@itwin/desktop-viewer-react"
+  },
+  {
+    "textKey": "home.links.customizeViewer",
+    "url":"https://developer.bentley.com/tutorials/itwin-viewer-hello-world/"    
+  }
+]
```
aruniverse, line 10: is it possible to open the sign in redirect using the same popups as you do for these?

packages/apps/desktop-viewer-test/.env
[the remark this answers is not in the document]
aruniverse: @calebmshafer are you aware of this being documented anywhere? i know i added a spot for this in the announcment i made, but i dont think thats the right spot

packages/apps/desktop-viewer-test/src/common/ViewerConfig.ts
```diff
@@ -33,3 +35,22 @@ export const viewerRpcs = [
...
+  LOCAL,
+}
+
+export interface ViewerFile {
+  displayName: string;
+  type: ViewerFileType;
+  path?: string;
+  iTwinId?: string;
+  iModelId?: string;
+}
```
aruniverse, line 51: should you add changeset id as well?

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  useEffect(() => {
+    if (initialized) {
+      void getUserSettings().then((userSettings) => {
+        setSettings(userSettings);
+      });
+    }
+  }, [initialized]);
+
+  const addRecentSnapshot = async (path: string) => {
```
aruniverse, line 37: should this and addRecentOnline be memoized with useCallback?

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const [searchParam, setSearchParam] = React.useState("");
+  const startSearch = React.useCallback(() => {
+    setSearchParam(searchValue);
+  }, [searchValue]);
+  const accessToken = useAccessToken();
+  const navigate = useNavigate();
+
+  return accessToken ? (
+    <>
+      <div className="itv-scrolling-container select-project">
```
aruniverse, line 60: minor issue here seeing dual scrollbars

packages/apps/desktop-viewer-test/src/frontend/components/home/Recents.tsx
```diff
@@ -0,0 +1,61 @@
...
+    await navigate(`/itwins/${iTwinId}/${iModelId}`);
+  };
+
+  return (
+    <nav>
+      {recents?.map((recent) => {
+        let displayValue = recent.displayName;
+        // limit display value to 25 chars
+        if (displayValue.length > 25) {
+          displayValue = `${displayValue.substr(0, 25)}...`;
```
aruniverse, line 37: should it be substr 0,22? (im including the ellipses, but thats probably overkill)

aruniverse (approved):
Overall looks great. Few minor comments, but really only 3 things:
1) need to confirm refresh token works, i think theres a bug with the auth layer but its probably in core (this can be done in a separate pr, i might take a look at this)
1) localization didnt work for me on the getting started pg, i need to look at that over again
1) theres a double scrollbar in the project select

The rest are minor, this looks pretty much good to go

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -156,7 +158,9 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
     ]);
 
     useEffect(() => {
-      void getViewState();
+      if (isMounted.current) {
+        void getViewState();
+      }
     }, [getViewState]);
```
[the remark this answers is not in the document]
kckst8: fixed

packages/modules/viewer-react/src/hooks/useBaseViewerInitializer.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { useEffect, useMemo, useState } from "react";
+
+import { ViewerProps } from "../components/BaseViewer";
+import { BaseInitializer } from "../services/BaseInitializer";
+import { getInitializationOptions, isEqual } from "../utilities";
+import { useIsMounted } from "./";
```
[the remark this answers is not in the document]
kckst8: fixed

packages/apps/desktop-viewer-test/public/locales/en-PSEUDO/iTwinViewer.json
kckst8: fixed

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
kckst8: updated

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.tsx
kckst8: fixed

packages/apps/desktop-viewer-test/src/frontend/components/home/Recents.tsx
kckst8: its really 28 chars....25 + ellipsis fits

packages/modules/viewer-react/src/hooks/useAccessToken.tsx
```diff
@@ -0,0 +1,36 @@
...
+    void getAccessToken();
+  }, [getAccessToken]);
+
+  useEffect(() => {
+    IModelApp?.authorizationClient?.onUserStateChanged.addListener((token) => {
+      if (isMounted.current) {
+        setAccessToken(token);
+      }
+    });
+  }, []);
```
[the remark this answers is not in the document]
kckst8: fixed

packages/apps/desktop-viewer-test/public/links.json
kckst8: Havent been able to find a way so far...core does not seem to offer an option, but perhaps a hack is possible

PR #47, opened 2021-09-10
CRA Web Template Refactor: Pull's out the viewer and Auth to separate files, removes header.

dustinlebsock:
The header is removed and sign-in is now automatically done on app start. The viewer will ask the user to sign-in if it fails. The viewer is pulled into a separate file as well as the authorization.


packages/modules/cra-template-web-viewer/template/src/App.tsx
```diff
@@ -44,121 +35,63 @@ const App: React.FC = () => {
...
+  BrowserAuthorizationCallbackHandler.handleSigninCallback(
+    process.env.IMJS_AUTH_CLIENT_REDIRECT_URI
+  ).then(() => {
+    client.signIn(new FrontendRequestContext());
+    history.push(
+      `?contextId=${sessionStorage.getItem(
+        "IMJS_CONTEXT_ID"
+      )}&iModelId=${sessionStorage.getItem("IMJS_IMODEL_ID")}`
+    );
+  });
```
dustinlebsock, line 90: Signing in on startup proved to be the most challenging task. Because of redirection, the client has no idea if it is signed in or not, and there is no easy method call to determine if there is an OIDC token currently in local storage. This will handle the callback, if a signin request has already been made, then attempt to signin again. With the second signin request, it will silent signing because the browser already has a token. Then the history is pushed because if it is pushed before this point, the callbackHandler is unable to get a token from the return url.

I couldn't find a more elegant solution, but this atleast works. Please give thoughts and opinions!

packages/modules/cra-template-web-viewer/template/src/Context.tsx
```diff
@@ -0,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { WebAuthorizationOptions } from "@itwin/web-viewer-react";
+import React from "react";
+
+export interface IModelJsConfig {
```
kckst8, line 8: let's remove "imodeljs..." from these names. Technically, there is no more imodeljs so, even though its referenced elsewhere, we don't want to add it in new places. ViewerContext might be more meaningful and will help to obscure the underlying tech. Also should rename the file to match the name of the context.

kckst8 (commented):
One minor comment about names but looks good otherwise! Hoping my desktop PR gets approved in the coming days at which point I will test this more thoroughly and get it merged

packages/modules/cra-template-web-viewer/template/src/history.ts
```diff
@@ -2,7 +2,11 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
aruniverse, line 2: dup headers

packages/modules/cra-template-web-viewer/template/src/history.ts
```diff
@@ -2,7 +2,11 @@
...
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
 
-import { createBrowserHistory } from "history";
+import createBrowserHistory from "history/createBrowserHistory";
```
aruniverse, line 10: any real benefit of changing this import?

packages/modules/cra-template-web-viewer/template/src/ViewerStartup.tsx
```diff
@@ -0,0 +1,67 @@
...
+} from "@bentley/imodeljs-frontend";
+import { Viewer } from "@itwin/web-viewer-react";
+import React, { useContext } from "react";
+
+import AuthorizationOptions from "./AuthorizationOptions";
+import IModelJsContext from "./Context";
+
+const ViewerStartup: React.FC = () => {
+  // Convert process var to window var
+  const context = useContext(IModelJsContext);
```
aruniverse, line 26: why not destructure here?

packages/modules/cra-template-web-viewer/template/src/App.tsx
```diff
@@ -44,121 +44,47 @@ const App: React.FC = () => {
...
-
-  const onLogoutClick = async () => {
-    setIsLoggingIn(false);
-    await IModelApp.authorizationClient?.signOut();
-    setIsAuthorized(false);
-  };
+  /** Grab the ContextId and IModelId from the URL parameters first, else the .ENV variables */
+  const urlParams = new URLSearchParams(window.location.search);
+  if (urlParams.has("contextId")) {
+    context.contextId = urlParams.get("contextId") as string;
```
aruniverse, line 50: eh instead of directly mutating context, if you changed whats in context to also have a setState callback that might be better

packages/modules/cra-template-web-viewer/template/src/App.tsx
```diff
@@ -3,31 +3,31 @@
...
-  const [isAuthorized, setIsAuthorized] = useState(
-    (IModelApp.authorizationClient?.hasSignedIn &&
-      IModelApp.authorizationClient?.isAuthorized) ||
-      false
-  );
-  const [isLoggingIn, setIsLoggingIn] = useState(false);
-  const [iModelId, setIModelId] = useState(process.env.IMJS_IMODEL_ID);
-  const [contextId, setContextId] = useState(process.env.IMJS_CONTEXT_ID);
+  // This will be the default context read from Context.tsx
+  const context = useContext(IModelJsContext);
```
aruniverse, line 28: you're calling useContext without wrapping the parent level component with the provider

packages/modules/cra-template-web-viewer/template/src/history.ts
dustinlebsock: with the change to the tsconfig
`"baseUrl": "src"`

It introduces a circular dependency. The baseUrl addition is useful for when we add our common/local modules to the template, its all based off the src folder rather than its location in the module.

packages/modules/cra-template-web-viewer/template/src/App.tsx
dustinlebsock: I was essentially using it as a global variable, which i thought was fine to limit the amount of code. Provider is added so it doesn't look like magic now.

packages/modules/cra-template-web-viewer/template/src/App.tsx
dustinlebsock: There are state variables now.

packages/modules/cra-template-web-viewer/template/src/ViewerStartup.tsx
dustinlebsock: destructured

packages/modules/cra-template-web-viewer/template/src/history.ts
dustinlebsock: removed

packages/modules/cra-template-web-viewer/template/src/Context.tsx
dustinlebsock: Renamed to ViewerContext.

packages/modules/cra-template-web-viewer/template/src/history.ts
kckst8: This causes this error/warning in the console:

image

Guess I'm not following how this import causes a circular dependency?

kckst8 (changes requested):

packages/modules/cra-template-web-viewer/template/src/App.tsx
```diff
@@ -2,32 +2,30 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-
+/*---------------------------------------------------------------------------------------------
```
kckst8, line 5: looks like a duplicate copyright header was generated here

kckst8 (approved):

PR #52, opened 2021-09-29
Fix issue with duplicate menu listeners

kckst8:
listeners were being registered multiple times in the hook and IpcApp.removeListener does not properly remove/unregister. Moved them to a once time initialization in itwinViewerApp


aruniverse (approved):
lgtm code wise, didnt pull to test but trust you

PR #68, opened 2021-11-08
change loading spinner to a bar and remove redundant text

kckst8:


packages/modules/viewer-react/src/components/iModel/IModelBusy.tsx
```diff
@@ -5,23 +5,23 @@
 
 import "./IModelBusy.scss";
 
-import { BodyText, LoadingSpinner, SpinnerSize } from "@bentley/ui-core";
+import { IModelApp } from "@bentley/imodeljs-frontend";
+import { ProgressBar } from "@bentley/ui-core";
```
aruniverse, line 9: could we just use the one from itwinui instead? like we're going to switch very soon

aruniverse: discussed internally, will just switch to itwinui in the next version

aruniverse (approved):

PR #70, opened 2021-11-15
Fix intellisense

aruniverse:
Fix intellisense by removing prop on ItwinViewerInitializerParams which allowed for anything to be passed in as a prop
This helped catch quite a few issues throughout the codebase

Added an example of a IModelBankClient that implements the new FrontendHubAccess interface which replaces the IModelClient on IModelApp

### DIFF
--- a/packages/apps/web-viewer-test/src/services/IModelBankFrontendHubAccess.ts
+++ b/packages/apps/web-viewer-test/src/services/IModelBankFrontendHubAccess.ts
@@ -1,0 +1,73 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import {
+  ChangeSet,
+  ChangeSetQuery,
+  IModelBankClient,
+  VersionQuery,
+} from "@bentley/imodelhub-client";
+import {
+  BentleyError,
+  BentleyStatus,
+  ChangesetId,
+  IModelVersion,
+} from "@itwin/core-common";
+import { FrontendHubAccess, IModelIdArg } from "@itwin/core-frontend";
+
+export class IModelBankFrontend implements FrontendHubAccess {
+  private _hubClient: IModelBankClient;
+  constructor(orchestratorUrl: string) {
+    this._hubClient = new IModelBankClient(orchestratorUrl, undefined);
+  }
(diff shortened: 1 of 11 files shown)


packages/modules/web-viewer-react/src/hooks/useWebViewerInitializer.tsx
```diff
@@ -15,7 +16,7 @@ import { WebViewerProps } from "../types";
 
 export const useWebViewerInitializer = (options?: WebViewerProps) => {
   const [webViewerInitOptions, setWebViewerInitOptions] =
-    useState<WebViewerProps>();
+    useState<ItwinViewerInitializerParams>();
```
aruniverse, line 19: @kckst8 I want to highlight this change and make sure you looked at this; i question it a bit. From my possibly flawed understanding, you were trying to memoize all the "base" viewer initializer props, but since the interface allowed for any kvp it didnt really pull those out. I was surprised i didnt have to update the useDesktopInitializer hook.

Here's the error which prompted me to change this.
image

I do think we probably want to memoize the authConfig too.

Want to get your thoughts, we can go discuss more on teams.

kckst8: This makes sense. It should be more specific. We should make the same change in the desktop initializer

packages/apps/web-viewer-test/src/services/IModelBankFrontendHubAccess.ts
```diff
@@ -0,0 +1,73 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import {
```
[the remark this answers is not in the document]
aruniverse: in 3.0 IModelApp.imodelClient has been replaced with IModelApp.hubAccess, [see](https://github.com/iTwin/itwinjs-core/blob/master/docs/changehistory/NextVersion.md#itwincore-frontend) [3.0](https://dev.itwinjs.org/reference/core-frontend/imodelapp/imodelappoptions/) [2.x](https://www.itwinjs.org/reference/imodeljs-frontend/imodelapp/imodelappoptions/)

I copied this client from the core full stack tests, but i do agree this default should be provided in one of the core client pkgs, and if not there in the viewer.

kckst8 (approved):

PR #77, opened 2021-12-01
Localization

aruniverse:
- add localization client ourselves in base initializer instead of having core default it for you
- fixed desktop localization (i18n file named the same so it never picked up the right strings)
- minor scss changes requested by bill


bsteinbk:
@aruniverse - Here is the code the I put in my-3.0-viewer to set UI defaults to Gino's latest recommendations. Maybe you could set these some else so users do not need to do this.

  const onIModelAppInit = () => {
    UiShowHideManager.setAutoHideUi(true);
    UiShowHideManager.setUseProximityOpacity(false);
  }

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
```diff
@@ -23,7 +23,12 @@ import {
 } from "./routes";
 
 const App = () => {
-  const initialized = useDesktopViewerInitializer();
+  (window as any).ITWIN_VIEWER_HOME = window.location.origin;
+
+  const initialized = useDesktopViewerInitializer({
```
kckst8, line 28: Why do we need additional names pace? Previously we made use of the itwinviewer ns that is registered in the base package.  Not possible?

packages/modules/cra-template-web-viewer/template/src/Header.tsx
```diff
@@ -26,6 +26,7 @@ export const Header: React.FC<HeaderProps> = ({
           className={styles.button}
           onClick={handleLogin}
           styleType={"cta"}
+          size={"small"}
```
kckst8, line 29: I thought we were removing the header in the cra template for this release and just signing in on startup?

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
aruniverse: Tbh not sure how localization worked in the previous gen or if it worked at all. There are strings shown on the home route that are defined in a   json file in the desktop app, but had the same file name as the one in the base package so there was a collision there, and it doesn't seem to merge them so we only get strings from the base viewer and not the desktop app. 

I thought this approach was better than moving the other strings into the base because now users can add their app strings too.

packages/modules/cra-template-web-viewer/template/src/Header.tsx
aruniverse: I wasn't aware of that, I wouldn't be opposed to it tho. Should that go in a separate pr then? I snuck the style changes in here cause I thought they were miniscule

kckst8: yes I'll add a PBI for it. I think it was actually removed with the changes to the template for sandbox export that we ended up reverting

kckst8 (approved):

PR #78, opened 2021-12-02
Desktop template with local briefcases

kckst8:
Some of this can be optimized a bit and the status bar extension and related components could be moved into the desktop viewer package to minimize code in the template. Those changes would involve the addition of iTwinUI and scss support in that package though, which is already scheduled for later. At that point we can revisit to avoid adding too much here prior to the 2.0 release


calebmshafer:
@kckst8 the download spinner doesn't seem to spin when it's in progress. I see the percentages being logged in the console but the spinner on the tile isn't doing anything.

calebmshafer:
One more nitpick, the current "Learn" section on the landing page has the links popping open a new Electron window instead of going directly to the system's default browser.

kckst8:
> @kckst8 the download spinner doesn't seem to spin when it's in progress. I see the percentages being logged in the console but the spinner on the tile isn't doing anything.
> 
> Edit: Also, clicking the close button on the test desktop viewer basically crashed the app. Not sure if that's part of the viewer code or just the template. I get,
> 
> Viewer-crash

Spinner should be fixed. When you say click the close button, do you mean the system close X to close the app? Does it not close for you?

kckst8:
> One more nitpick, the current "Learn" section on the landing page has the links popping open a new Electron window instead of going directly to the system's default browser.

I changed the behavior, but this is the default for links in Electron apps. It was basically a hack to change it. Seeing as this is an app template, we should decide if that should be our standard (though I did add a code comment to remove if not desired)

calebmshafer:
> Spinner should be fixed.

cool, I'll pull and give it a try.

> When you say click the close button, do you mean the system close X to close the app? Does it not close for you?

Two things, the close X button doesn't work but I am in development mode so let me verify if it works in a prod build. The second one is when I go to "File > Close", this is the one that throws that error I posted above and then leaves it in a bad state.

calebmshafer:
@kckst8 I'm now getting a different error with the ViewCreator when I switch from a Briefcase (which opens correctly), click on the "File > Open", and choose a Snapshot iModel on disk. Here's the error, let me know if you need more details.

viewCreatorError

calebmshafer:
> > One more nitpick, the current "Learn" section on the landing page has the links popping open a new Electron window instead of going directly to the system's default browser.
> 
> I changed the behavior, but this is the default for links in Electron apps. It was basically a hack to change it. Seeing as this is an app template, we should decide if that should be our standard (though I did add a code comment to remove if not desired)

@kckst8 Ah I guess if that's the default behavior we should probably leave it... Though not a huge fan of it :)

packages/apps/desktop-viewer-test/src/frontend/components/home/Recents.tsx
```diff
@@ -36,25 +32,11 @@ export const Recents = () => {
...
-                onClick={() => openRemote(recent.iTwinId, recent.iModelId)}
-              >
-                {displayValue}
-              </span>
-            );
-          default:
-            return null;
-        }
+        return (
+          <span key={recent.path} onClick={() => openFile(recent.path)}>
```
aruniverse, line 36: i wouldve assumed path is unique, but got a warning in the console about list needed unique keys

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/BriefcaseStatus.tsx
```diff
@@ -0,0 +1,113 @@
...
+        return <SvgSync onClick={onMergeClick} />;
+      case ModelStatus.UPTODATE:
+        return <SvgStatusSuccess />;
+      case ModelStatus.ERROR:
+        return <SvgStatusError />;
+      case ModelStatus.ONLINE:
+        return <SvgDownload onClick={onDownloadClick} />;
+      default:
+        // no UI if the status isn't handled
+        return null;
```
aruniverse, line 104: deplist is missing a few, mergeProgess specifically wont actually update the spinner

packages/apps/desktop-viewer-test/src/frontend/extensions/IModelMergeStatusBarItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { BriefcaseStatus } from "../components/modelSelector";
+
+const MergeStatusBarItem = () => {
+  const [mergeStatus, setMergeStatus] = useState<ModelStatus>();
+  const [connection, setConnection] = useState<BriefcaseConnection>();
+  const accessToken = useAccessToken();
+  const connectivityStatus = useConnectivity();
+
+  const onMergeClick = async () => {
+    setMergeStatus(ModelStatus.MERGING);
```
aruniverse, line 35: couldnt you just call getLatestChanges?

packages/apps/desktop-viewer-test/src/frontend/extensions/IModelMergeStatusBarItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const onLoginClick = async () => {
+    await IModelApp.authorizationClient?.signIn();
+  };
+
+  const getLatestChangesets = useCallback(async () => {
+    if (connection) {
+      try {
+        await connection.pullAndMergeChanges();
+        setMergeStatus(ModelStatus.UPTODATE);
+        IModelApp.viewManager.refreshForModifiedModels(undefined);
```
aruniverse, line 47: Is this actually even possible? Merging a model while its open? Ive had no lcukw ith it so far, fails to apply changesets.

The sqlite file is open for read, so how would it apply/write any new changes?

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
```diff
@@ -5,34 +5,194 @@
...
+  useEffect(() => {
+    if (!briefcase) {
+      void getBriefcase();
+    }
+    return () => {
+      if (briefcase) {
+        void briefcase.close();
+      }
+    };
+  }, []);
```
aruniverse, line 123: is this missing a dep on briefcase?

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
```diff
@@ -5,34 +5,194 @@
...
+        setStatus(ModelStatus.ONLINE);
+      }
+      return fileName;
+    } catch {
+      setStatus(ModelStatus.ERROR);
+    }
+  }, []);
+
+  const mergeChanges = useCallback(() => {
+    setStatus(ModelStatus.MERGING);
```
aruniverse, line 98: same comments as the status bar item, couldnt this call getLatestChangesets?

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
```diff
@@ -5,34 +5,194 @@
...
+
+  const { progress, doDownload } = useDownload(
+    iModel.id,
+    iModel.name ?? iModel.id,
+    iModel.projectId ?? ""
+  );
+
+  const getLatestChangesets = useCallback(async () => {
+    if (briefcase) {
+      await briefcase.pullAndMergeChanges();
```
aruniverse, line 78: how are you debugging the backend? specifically backend core

aruniverse (commented):
Overall most of the code looks good. Merging changes have been very hit or miss for me. Trying to debug and figure out whats failing

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/BriefcaseStatus.tsx
aruniverse: gah caleb beat me to it

aruniverse:
Nit, i worry about calling that one dir `extensions`

packages/apps/desktop-viewer-test/src/frontend/components/App.tsx
```diff
@@ -3,49 +3,61 @@
...
       void getUserSettings().then((userSettings) => {
         setSettings(userSettings);
       });
     }
+    return () => {
+      if (connectivityListener) {
+        connectivityListener();
+      }
+    };
   }, [initialized]);
```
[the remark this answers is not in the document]
kckst8: I don't think so. This is where we set connectivityListener, so it would cause an un-necessary effect....its only being referenced in the return statement (to basically remove the listener) under the assumption that it is not being set anywhere else.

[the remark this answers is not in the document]
kckst8: its saved so that it can be removed....otherwise it would be a potential memory leak

packages/apps/desktop-viewer-test/src/frontend/components/home/Recents.tsx
kckst8: probably because the tooling cant guarantee uniqueness based on its type......we are only every persisting a single path once so it should be, but I'll look at alternatives

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
kckst8: there is a debugger configured in launch.json for the backend ( itwin-viewer-desktop -> BACKEND). Typically I run  npm start and then start the debugger. At that point breakpoints, etc should work. You can step into core or add breakpoints in the node_modules. Not great but sufficient.

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
kckst8: it could be since we have UI tied to the status state, this ensures that updates occur in the proper timeframe. For example, if we set the status to merging, and then call getLatestChangesets, and then change the status again immediately after, the UI would not update appropriately (or timely I should say). This ensures that getLatestChangesets executes once the spinner is rendered

packages/apps/desktop-viewer-test/src/frontend/extensions/IModelMergeStatusBarItem.tsx
kckst8: same comment as above

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
aruniverse: Ok I was hoping there'd be a better solution than going through node_modules to set breakpoints, but I'll try that out.

packages/apps/desktop-viewer-test/src/frontend/extensions/IModelMergeStatusBarItem.tsx
kckst8: the briefcase connection is opened in readonly state (see openLocalIModel in IModelService), which means you can "pull" changes, but not push. You definitely cannot merge changes back to the model. That is out of scope here. Merge maybe isn't the best word. Maybe just pulling?

What are you attempting here? The proper way to test this is to open a model, then sync new changes to the model, and then click merge. The new changes should be evident in the local briefcase. I havent had any issues with that workflow so let me know if I am missing something

aruniverse: Downloaded model
Sync changes via demo portal
Restarted app, opened the downloaded model before merging
The status bar item lets me merge (which i dont think it should) so i click it, fails to merge/pull
image
image
image

Fails to close the bim file when its open in the viewer, which makes sense.  I dont think it should actually should getLatestChnageset when youre in the viewer, just alert the user to go back and merge from the Select pg again

aruniverse: I then navigated back to the imodel selection pg (which i wouldve assumed closed the connection), and then tried merging  again but that failed with the same applying changeset error.

I opened another model, went back to merge again still failed. Not sure whats doing on. The only error i see in the log is BE_SQLITE_BUSY

aruniverse: Restarting the viewer, and merging from the select pg without ever opening the model is fine, but if you open it and then try merging it will def fail in the native layer when trying to apply the changesets

kckst8: Ok thanks.....I havent been able to replicate this so I'm going to attempt on Windows.

kckst8:
> Nit, i worry about calling that one dir `extensions`

My thought is that UIProviders === extensions for the time being, and when we move to the extensions API this item will be converted to use that. So, it didn't seem useful to me to have to rename and explain the differences later as opposed to just changing the underlying code for the same feature with little interruption/explanation needed. 

Open to changing if enough feel otherwise

kckst8:
> @kckst8 I'm now getting a different error with the ViewCreator when I switch from a Briefcase (which opens correctly), click on the "File > Open", and choose a Snapshot iModel on disk. Here's the error, let me know if you need more details.
> 
> viewCreatorError

this should be resolved in my latest push

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectIModel.tsx
kckst8: fixed

packages/apps/desktop-viewer-test/src/frontend/extensions/IModelMergeStatusBarItem.tsx
kckst8: As discussed, this appears to be a bug in core. Essentially the iModel DB allows pulling changes in a read-only connection. To do so, it temporarily re-opens as writable, applies changes, and then attempts to re-open as readonly again. The last part is what is failing. It appears to have been addressed in 3.0, so will re-test again once merged into next. In the meantime, I made some other changes that will hopefully help with displaying the changed model in the viewport. Please verify and let me know if that still does not occur

aruniverse: Working for me now.

packages/apps/desktop-viewer-test/src/frontend/extensions/IModelMergeStatusBarItem.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { useSelector } from "react-redux";
+
+import { BriefcaseStatus } from "../components/modelSelector";
+
+const MergeStatusBarItem = () => {
+  const [mergeStatus, setMergeStatus] = useState<ModelStatus>();
+  const [connection, setConnection] = useState<BriefcaseConnection>();
+  const accessToken = useAccessToken();
+  const connectivityStatus = useConnectivity();
+  const iModelConnection = useSelector((state: any) => {
```
aruniverse, line 39: guess you couldnt use the useActiveIModelConnection hook because thats dependent on the view maanger which will be stale/out of synch, so you needed to access it like so

aruniverse (approved):
lgtm, everythings working as expected now

packages/apps/desktop-viewer-test/src/frontend/extensions/IModelMergeStatusBarItem.tsx
kckst8: I would prefer to use that hook but didn't get the results that I expected. However, that was pre-refactor. I'll try again before merging

kckst8: Works as expected now. I updated. Great catch.

PR #79, opened 2021-12-07
Update the desktop CRA template with local briefcase support

kckst8:
Adds support for local briefcases (the same code from the previously reviewed test app)


packages/modules/cra-template-desktop-viewer/template.json
```diff
@@ -34,10 +34,10 @@
       "@bentley/ui-framework": "^2.14.0",
       "@bentley/ui-ninezone": "^2.14.0",
       "@bentley/webgl-compatibility": "^2.14.0",
-      "@itwin/desktop-viewer-react": "^1.0.16",
+      "@itwin/desktop-viewer-react": "^1.2.0",
```
aruniverse, line 37: i think moving forward (after 3.0) we can just use the `latest` dist tag here

packages/modules/cra-template-desktop-viewer/template/src/react-app-env.d.ts
```diff
@@ -0,0 +1,6 @@
+/*---------------------------------------------------------------------------------------------
```
aruniverse, line 1: i dont think you need this file, should get auto generated when you create a new app using our cra template and fork

aruniverse (approved):

packages/modules/cra-template-desktop-viewer/template.json
kckst8: agree, as long as we're still using the next tag for prerelease

packages/modules/cra-template-desktop-viewer/template/src/react-app-env.d.ts
kckst8: Mistake copy. I deleted

PR #80, opened 2021-12-08
Add RealityClient to Viewer

aruniverse:
Core willl not create a reality data client for you
Update .env file to add the new scope needed


packages/modules/viewer-react/src/services/BaseInitializer.ts
```diff
@@ -260,5 +261,6 @@ export const getIModelAppOptions = (
         options?.i18nUrlTemplate ??
         (viewerHome && `${viewerHome}/locales/{{lng}}/{{ns}}.json`),
     }),
+    realityDataAccess: new RealityDataAccessClient(),
```
calebmshafer, line 264: Do we want to always initialize one for users and not allow them to configure it?  I'm a little concerned about them not being able to configure it.

aruniverse: Typically we've been adding ways for users to override our defaults only after they request it. Im not opposed to exposing it. I defer to @kckst8

kckst8: @calebmshafer do you expect this to be a common request? Keeping in mind that one of the goals of the Viewer is to obscure large amounts of boilerplate and excess configurability

packages/modules/web-viewer-react/package.json
```diff
@@ -90,6 +91,7 @@
     "@itwin/presentation-common": "rc",
     "@itwin/presentation-components": "rc",
     "@itwin/presentation-frontend": "rc",
+    "@itwin/reality-data-client": "latest",
```
kckst8, line 94: Is this temporary until release? I don't think we should use latest in package.json. We need a specific minimum that is supported. Template.json in the CRA templates is probably ok to use latest

aruniverse: Yea i was going to go back and update all the deps to an actual version once theyre stable and officially released

aruniverse:
@calebmshafer do we expect a lot of users to have reality-data? I was originally thinking we just add an option for proving your own realityDataAccess in the viewer props and call it a day. This way if you don't have reality data you dont have to pull that into your bundle for no reason, dont need any additional scopes, etc

calebmshafer:
> > @calebmshafer do we expect a lot of users to have reality-data? I was originally thinking we just add an option for proving your own realityDataAccess in the viewer props and call it a day. This way if you don't have reality data you dont have to pull that into your bundle for no reason, dont need any additional scopes, etc
> 
> If this is the case, my preference would be a boolean prop (assuming we don't expect many custom reality data clients). Asking users to create their own clients to meet common use cases goes against our goal of reducing boilerplate and being a quick start tool

@aruniverse @kckst8 I think this should be on by default since reality data is a core part of our iTwin Platform and a major part of digital twins. If people don't want the dependency they can at that point fork the Viewer. In terms of the things we pull into the Viewer this one is fairly low on my issues list :)

One thing we should do is look at how we're going to expose the ability to configure the Bing Maps or other such keys on `iModelApp.startup`. Is that already planned, or maybe in there (haven't thoroughly looked)?

aruniverse:
> One thing we should do is look at how we're going to expose the ability to configure the Bing Maps or other such keys on `iModelApp.startup`. Is that already planned, or maybe in there (haven't thoroughly looked)?

Yea i was thinking about this the other day, would honestly be easy if we exposed everything from IModelAppOptions in the viewer, but that goes against the minimal options philosophy we've taken so far. I think we need to add a pbi for the BingMaps and MapLayersOptions one

kckst8 (approved):
LGTM

PR #83, opened 2021-12-10
Bump to core rc 2 (dev146)

aruniverse:
Bump to core rc 2 (dev.146)
Fix memoization of initialization props
Use FrontendDevTools in desktop test


packages/modules/viewer-react/src/utilities/index.ts
```diff
@@ -15,7 +15,7 @@ export * from "./MakeCancellable";
  * @returns
  */
 export const isObject = (item: any) => {
-  return Object.prototype.toString.call(item) === "[object Object]";
+  return item !== null && typeof item === "object";
```
kckst8, line 18: this will return true for arrays, which we don't want

packages/modules/desktop-viewer-react/src/hooks/useDesktopViewerInitializer.tsx
```diff
@@ -16,7 +16,7 @@ import type { DesktopViewerProps } from "../types";
 
 export const useDesktopViewerInitializer = (options?: DesktopViewerProps) => {
   const [desktopViewerInitOptions, setDesktopViewerInitOptions] =
-    useState<ItwinViewerInitializerParams>();
+    useState<Partial<ItwinViewerInitializerParams>>();
```
kckst8, line 19: why partial here?

packages/modules/viewer-react/src/utilities/index.ts
```diff
@@ -62,12 +62,17 @@ export const isEqual = (a?: any, b?: any) => {
...
   if (options) {
-    for (const key in iTwinViewerInitializerParamList) {
-      (initOptions as any)[key] = (options as any)[key];
+    for (const key of iTwinViewerInitializerParamList) {
+      if ((options as any)[key]) {
+        (initOptions as any)[key] = (options as any)[key];
+      }
     }
   }
+  console.log("iTwin.js initialization options: ", initOptions);
```
kckst8, line 76: should we be logging this?

packages/modules/viewer-react/src/services/BaseInitializer.ts
```diff
@@ -260,5 +263,6 @@ export const getIModelAppOptions = (
         options?.i18nUrlTemplate ??
         (viewerHome && `${viewerHome}/locales/{{lng}}/{{ns}}.json`),
     }),
+    publicPath: `${viewerHome}/` ?? "",
```
aruniverse, line 266:
```suggestion
    publicPath: viewerHome ? `${viewerHome}/` :  "",
```

packages/apps/desktop-viewer-test/src/frontend/components/viewer/ModelViewer.tsx
```diff
@@ -44,16 +45,17 @@ export const ModelViewer = ({
     });
   }, []);
 
-  const onIModelAppInitialized = useMemo(() => {
-    console.log("Initialized!!");
+  const onIModelAppInitialized = useCallback(async () => {
+    await FrontendDevTools.initialize();
```
aruniverse, line 49: For some reason, why i try to initialize this, IModelApp.localization is undefined, seems to be using the esm ver here, but the rest of the app uses cjs... not sure whats going on

packages/modules/viewer-react/src/utilities/index.ts
aruniverse: good catch

packages/modules/viewer-react/src/utilities/index.ts
aruniverse: just for testing, removed

packages/modules/desktop-viewer-react/src/hooks/useDesktopViewerInitializer.tsx
aruniverse: good point, makes no difference here since all the props are optional, removed

PR #85, opened 2021-12-13
Add map-layers-options to initializer props

aruniverse:


aruniverse:
Discsusing the addition of the axios peer dep that slipped in during my last pr in this thread https://github.com/iTwin/reality-data-client/pull/29. Axios should not be a peer dep imo, and could probably just use fetch instead. Will update other pkgs as needed depending on how that connvo goes. Added axios as  dep in the web-test

kckst8 (approved):

PR #86, opened 2021-12-13
Performance Monitoring

kckst8:
Add performance measures and the ability to log them in AI


packages/modules/viewer-react/src/services/telemetry/Performance.ts
```diff
@@ -0,0 +1,95 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { trackViewerMetric, viewerAI } from "./TelemetryService";
+
+export type PerformanceMeasures =
+  | "ViewerInitialized"
+  | "BaseViewerInitialized"
+  | "IModelConnected"
+  | "ViewStateCreated"
+  | "TileTreesLoaded";
```
aruniverse, line 13: Going to be an interesting measure to track. Would probably want to check with paul or mark or the current perf tests guys how theyre handling this, if you havent already.

packages/modules/viewer-react/src/services/telemetry/Performance.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (
+      performance.getEntriesByName(startMark).length > 0 &&
+      performance.getEntriesByName(endMark).length > 0
+    ) {
+      performance.measure(measureName, startMark, endMark);
+    }
+  }
+
+  static async logMetric(measureName: PerformanceMeasures) {
+    if (!viewerAI.initialized) {
```
aruniverse, line 69: probabluy want to check if perf tracking is enabled here too

packages/modules/viewer-react/src/services/BaseInitializer.ts
```diff
@@ -31,10 +31,11 @@ import type { ElectronRendererAuthorization } from "@itwin/electron-authorizatio
 import { PresentationRpcInterface } from "@itwin/presentation-common";
 import { Presentation } from "@itwin/presentation-frontend";
 
+import { ViewerPerformance } from "..";
```
aruniverse, line 34: prefer not to import from internal barrel

packages/modules/desktop-viewer-react/src/services/Initializer.ts
```diff
@@ -45,9 +47,15 @@ export class DesktopInitializer {
       this._initializing = true;
 
       const cancellable = makeCancellable(function* () {
+        ViewerPerformance.enable(options?.enablePerformanceMonitors);
+        ViewerPerformance.addMark("ViewerStarting");
         const additionalRpcInterfaces = options?.additionalRpcInterfaces ?? [];
         additionalRpcInterfaces.push(SnapshotIModelRpcInterface);
 
+        if (!options) {
```
aruniverse, line 55: is this chunk neded?

aruniverse (commented):
sorry, started reviewing before i noticed it said draft.

packages/modules/viewer-react/src/utilities/index.ts
```diff
@@ -63,11 +63,11 @@ export const isEqual = (a?: any, b?: any) => {
  * @returns
  */
 export const getInitializationOptions = (options?: ViewerProps) => {
-  const initOptions: ItwinViewerInitializerParams = {};
+  const initOptions = {};
   if (options) {
     for (const key in iTwinViewerInitializerParamList) {
```
aruniverse, line 68: use key of on L68!

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -31,10 +31,11 @@ import { withAITracking } from "@microsoft/applicationinsights-react-js";
 import React, { useCallback, useEffect, useState } from "react";
 import { Provider } from "react-redux";
 
+import { ViewerPerformance } from "../..";
```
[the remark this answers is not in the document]
kckst8: updated. Still using "a" barrel, but only the telemetry services barrel so it should not cause un-desirable effects

packages/modules/viewer-react/src/services/telemetry/Performance.ts
kckst8: this was the best method to track when tiles are loaded when I implemented, but I'll double check with the display team and update if necessary

packages/modules/viewer-react/src/services/telemetry/Performance.ts
kckst8: I changed this to private and only called from the internal methods. We shouldn't need/allow logging without adding, but we should allow adding without logging so I left that one public

packages/modules/viewer-react/src/services/BaseInitializer.ts
kckst8: updated

packages/modules/desktop-viewer-react/src/services/Initializer.ts
kckst8: it is as the code stands since we spread the options below. I went back and forth between making options not optional and leaving optional and assuming false here. My rationale is that it is still a required prop of the component that calls this function, but we don't have to change the function and caller this way. Definitely open to discussion here. This is the first required initialization parameter that we've added.

packages/modules/viewer-react/src/utilities/index.ts
kckst8: I added this change to this PR

packages/modules/desktop-viewer-react/README.md
```diff
@@ -47,6 +47,16 @@ export const MyViewerComponent = () => {
 
 ### Props
 
+#### Required
+
+- `enablePerformanceMonitors` - Enable reporting of data from timed events in the iTwin Viewer in order to aid in future performance optimizations. These are the metrics that will be collected and logged to the browser's performance timeline as well as to Azure Application Insights:
+  - Duration of startup to the initialization of iTwin.js services
+  - Duration of startup to the establishment of a connection to the iModel
+  - Duration of startup to the creation of a view state for the iModel
+  - Duration of startup until the last tile is loaded and rendered for the iModel view
```
aruniverse, line 56: As I understand it, there are an "infinite" number of tiles, eg changing the zoom and LoD will just spawn more tiles to be generated/fetched.  Using the term "last tile" here can be misleading imo.

kckst8: this mark is added in a one time callback when the view opens though, so it technically is the last tile loaded initially (i.e. it will not mark again when resizing or changing the view)

packages/modules/viewer-react/src/services/telemetry/Performance.ts
```diff
@@ -0,0 +1,95 @@
...
+      case "qa-":
+        return "bc9fee8c-d537-4892-b760-750392c531be";
+      default:
+        return "76ebaa63-f57e-4955-aedf-43e2741724ec";
+    }
+  }
+
+  private static async _logMetric(measureName: PerformanceMeasures) {
+    if (!viewerAI.initialized) {
+      await viewerAI.initialize(this._getAiKey());
```
aruniverse, line 39: you can drop the async and await modifiers, viewer ai init is synchronous

packages/modules/web-viewer-react/src/services/Initializer.ts
```diff
@@ -85,20 +87,27 @@ export class WebInitializer {
       console.log("starting web viewer");
       this._initializing = true;
       const cancellable = makeCancellable(function* () {
+        ViewerPerformance.enable(options?.enablePerformanceMonitors);
```
aruniverse, line 90: options is required, as of the auth change pr, can drop the ?.

packages/modules/desktop-viewer-react/src/services/Initializer.ts
aruniverse: i _think_ we can make options required in the desktop viewer too now, which could help us eliminate this chunk

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
```diff
@@ -72,6 +73,12 @@ export class ViewCreator3d extends ViewCreator {
               const start = new Date();
               const intvl = setInterval(() => {
                 if (viewPort.areAllTileTreesLoaded) {
+                  ViewerPerformance.addMark("TilesLoaded");
```
aruniverse, line 76: how helpful will metrics like this be, if we're not tracking with it the number of tiles?

aruniverse:
Changes look fine. Just want to note, depending on your adblocker, these tracking reqs will fail

aruniverse (approved):
Changes look fine. Just want to note, depending on your adblocker, these tracking reqs will fail

packages/modules/desktop-viewer-react/README.md
calebmshafer: maybe add a comment here to say "for the initial iModel view"?

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
kckst8: the metrics in general are debatable due to so many potential factors. The ask here was to attempt to capture the time for the viewport to be "usable". I'm not sure of a better place for that but certainly open to suggestions.

aruniverse: This spot is fine imo, i just think we should also log the number of tiles with it, otherwise, it really is meaningless, and why bother. You should be able to use vp.numReadyTiles

aruniverse: https://github.com/iTwin/itwinjs-core/blob/master/test-apps/display-test-app/src/frontend/TileLoadIndicator.ts

the `onFinishRender` looks promising

packages/modules/desktop-viewer-react/src/services/Initializer.ts
kckst8: fixed

packages/modules/viewer-react/src/services/telemetry/Performance.ts
kckst8: fixed

packages/modules/web-viewer-react/src/services/Initializer.ts
kckst8: fixed

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
kckst8: I saw that but its fired in the render loop, which is oddly prior to tile requests: https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/IModelApp.ts#L495

packages/modules/desktop-viewer-react/README.md
kckst8: updated

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
aruniverse: Nice catch... that is weird. 

So the current event looks fine, thoughts on adding metrics like numReadyTiles?

PR #87, opened 2021-12-13
switch to new hub client

aruniverse:
- switch to new hub clients and update deps
  - requires `imodels:read` scope
- update reality data pkg to pickup fix for axios peer dep, and actually hit other envs

### DIFF
--- a/packages/modules/viewer-react/src/services/iModel/IModelService.ts
+++ b/packages/modules/viewer-react/src/services/iModel/IModelService.ts
@@ -17,14 +17,13 @@ const getVersion = async (
   if (changeSetId) {
     return IModelVersion.asOfChangeSet(changeSetId);
   }
-  const { authorizationClient, hubAccess } = IModelApp;
-  const token = await authorizationClient?.getAccessToken();
 
-  if (token && hubAccess) {
+  const accessToken = await IModelApp.authorizationClient?.getAccessToken();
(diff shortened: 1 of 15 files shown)


packages/modules/viewer-react/src/services/iModel/IModelService.ts
```diff
@@ -19,20 +17,15 @@ const getVersion = async (
...
-  const token = await IModelApp.authorizationClient?.getAccessToken();
-  if (token) {
+
+  const accessToken = await IModelApp.authorizationClient?.getAccessToken();
+  if (accessToken && IModelApp.hubAccess) {
     try {
-      const hubClient = new IModelHubClient(); // TODO 3.0 - this needs to be configurable to support iTwin Stack
-      const namedVersions = await hubClient.versions.get(
-        token,
+      const changeset = await IModelApp.hubAccess.getChangesetFromNamedVersion({
```
kckst8, line 24: wouldnt this require a named version name? I thought we were going to query for that on the client for now?

aruniverse: if the named version param is undefined, itll fetch the latest named ver. And i was able to query it in on the client (in the FrontendHubAccess client)

kckst8 (approved):

PR #89, opened 2021-12-15
Simplify viewer auth requirement

aruniverse:
Viewer now requires an auth client that implements ViewerAuthorizationClient, which extends the basic AuthorizationClient from core and adds an event for onAccessTokenChanged.


packages/modules/viewer-react/src/components/BaseBlankViewer.tsx
```diff
@@ -51,31 +50,23 @@ export const BaseBlankViewer: React.FC<BlankViewerProps> = ({
...
-        false
-    );
-    ViewerAuthorization.client?.onAccessTokenChanged.addListener(() => {
-      setAuthorized(
-        (ViewerAuthorization.client?.hasSignedIn &&
-          ViewerAuthorization.client?.isAuthorized) ||
-          false
-      );
-    });
+    const removeListener =
```
kckst8, line 53: should we just use the hook in the base components?

kckst8 (commented):
looks good so far. Still not sure about the CDN case. I need to verify that on my end. Few comments

aruniverse:
just pushing up to switch machines tomorrow. everything compiles. desktop app doesnt seem to run, web-test needs to update to the changes here

kckst8:
updated the CDN viewer test app to react to these changes. All looks good on that end when you are ready to publish

aruniverse:
I _think_ the pr is good to go. Let me know if you have any concerns over this commit: https://github.com/iTwin/viewer/pull/89/commits/a4668870fd46fcdcbe3b1cb0e435390ea65a556a. Will have to come back to cleanup the web-test app

kckst8 (approved):

PR #92, opened 2021-12-21
Merge master into next

kckst8:
NOTE: There is a bug in the backend hub access client that prevents the download of some iModels. It was fixed in this PR: [@itwin/imodels-access-backend. ](https://github.com/iTwin/imodels-clients/pull/68/files). Will need to upgrade that package once it the fix is published.


packages/modules/desktop-viewer-react/src/services/BriefcaseServices.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Guid } from "@itwin/core-bentley";
+import { IModelVersion } from "@itwin/core-common";
+import type { BriefcaseConnection } from "@itwin/core-frontend";
+import { CheckpointConnection } from "@itwin/core-frontend";
+
+import { ModelStatus } from "..";
```
aruniverse, line 11: prefer importing from "../types"

packages/apps/desktop-viewer-test/src/common/ViewerConfig.ts
```diff
@@ -36,16 +46,9 @@ export const viewerRpcs = [
...
-  ONLINE,
-  SNAPSHOT,
-  LOCAL,
-}
-
 export interface ViewerFile {
   displayName: string;
-  type: ViewerFileType;
-  path?: string;
+  path: string;
```
aruniverse, line 51: just want to confirm this is required now

packages/apps/desktop-viewer-test/src/frontend/components/routes/ViewerRoute.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }, [location?.state]);
+
+  const onIModelAppInitialized = useMemo(() => {
+    console.log("iTwin.js Initialized!");
+  }, []);
+
+  return filePath ? (
+    <Viewer
+      snapshotPath={filePath}
+      onIModelAppInit={onIModelAppInitialized as any}
```
aruniverse, line 37: did you need to cast it as any? was tsc complaining?

packages/apps/desktop-viewer-test/src/frontend/hooks/useDownload.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const addRecent = useCallback(async (fileName: string) => {
+    await userSettings.addRecent(fileName, iModelName, iTwinId, iModelId);
+  }, []);
+
+  const doDownload = useCallback(async () => {
+    const fileName = await ITwinViewerApp.saveBriefcase(iModelName);
+    if (fileName) {
+      const req = await NativeApp.requestDownloadBriefcase(
+        iTwinId as string,
```
aruniverse, line 30: cast should be uneccessary here, its already of type string

aruniverse (approved):
did a quick pass, changes look fine, (didnt look at the cra-desktop template but comments made in the desktop test app should apply)

desktop test app worked fine, ran into the issue downloading some modle,s but should be fixed with the changes you mentioned

packages/apps/desktop-viewer-test/src/frontend/components/routes/ViewerRoute.tsx
kckst8: its because it's memoized...but we should really require it to be. Maybe we should change the type accordingly. I'll look into it

aruniverse: Thinking about it, should probably use useCallback instead of useMemo

kckst8: yeah you're right.....I dont know why useMemo returns void, but I guess thats to force usecallback for functions

packages/apps/desktop-viewer-test/src/frontend/hooks/useDownload.tsx
kckst8: fixed

packages/apps/desktop-viewer-test/src/frontend/components/routes/ViewerRoute.tsx
kckst8: fixed

packages/modules/desktop-viewer-react/src/services/BriefcaseServices.ts
kckst8: fixed

PR #95, opened 2022-01-04
re-add tree-widget

aruniverse:


common/config/rush/common-versions.json
```diff
@@ -3,6 +3,7 @@
 
   "preferredVersions": {
     "lint-staged": "^10.2.11",
-    "immer": "^9.0.6"
+    "immer": "^9.0.6",
+    "react-error-overlay": "6.0.9"
```
kckst8, line 7: What caused the need for this?

kckst8 (approved):
Didn't test but changes lgtm

common/config/rush/common-versions.json
aruniverse: Kept running into:  Uncaught ReferenceError: process is not defined. This can be resolved by pinning down the version of react-error-overlay to 6.0.9. create-react-app/issues/11773.

PR #96, opened 2022-01-05
bump imodels and auth clients to latest

aruniverse:


packages/apps/desktop-viewer-test/src/backend/ViewerHandler.ts
```diff
@@ -104,12 +104,13 @@ class ViewerHandler extends IpcHandler implements ViewerIpc {
...
         const issuerUrl = getAppEnvVar("ISSUER_URL");
 
-        const authClient = await ElectronMainAuthorization.create({
+        const authClient = new ElectronMainAuthorization({
           clientId,
           scope,
           redirectUri: redirectUri || undefined, // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
           issuerUrl: issuerUrl || undefined, // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
         });
+        await authClient.signInSilent();
```
kckst8, line 113: this is odd. Are we required to sign in every time we are online? If so, should we catch errors here?

aruniverse: See https://github.com/iTwin/auth-clients/pull/47. The previous snippet would also attempt a silent sign-in as well. This shouldn't throw so i don't think we need to wrap it.

I don't think we need to sign in every time we are online, only once when we try to access the hub and the token is expired or is undefined. But if youre already online and connected (signed in with a valid access token) then i see no harm.

kckst8 (approved):

kckst8 (commented):
Be sure to make the same changes in the desktop CRA template that were made in the test app

aruniverse:
> Be sure to make the same changes in the desktop CRA template that were made in the test app

thx always forget this on my own prs 😅

PR #97, opened 2022-01-06
rename authConfig and remove unnecessary route from test app

kckst8:
some general cleanup as well


aruniverse (approved):
thx! been meaning to rename it

PR #98, opened 2022-01-06
re-add property grid

aruniverse:


aruniverse:
Theres some minor styling issues with extra scrollbars, need to talk to core about this and fix in vcr pkgs, but everything should be functional as before

kckst8 (approved):

PR #99, opened 2022-01-07
Update backend name

aruniverse:
Hold PR until gpb is deployed with new name

### DIFF
--- a/packages/modules/web-viewer-react/src/services/Initializer.ts
+++ b/packages/modules/web-viewer-react/src/services/Initializer.ts
@@ -40,7 +40,7 @@ const getHostedConnectionInfo = async (
     };
   } else {
     return {
-      info: { title: "visualization", version: "v3.0" },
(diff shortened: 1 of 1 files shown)


packages/modules/web-viewer-react/src/services/Initializer.ts
```diff
@@ -15,12 +15,12 @@ import {
 
 import type { IModelBackendOptions, WebViewerProps } from "../types";
 
-const getHostedConnectionInfo = async (
```
kckst8, line 18: lol. wonder why this was async? had to be a good reason at some point

kckst8 (approved):

packages/modules/web-viewer-react/src/services/Initializer.ts
aruniverse: in 2.x we probably got the orch url using buddi and needed it to be async

PR #102, opened 2022-01-11
remove identity scopes

kckst8:
remove identity scopes as they are no longer necessary


kckst8:
> could you update nextversion.md? also, is this supposed to be replaced with the `users` scope? or not needed at all?

updated NextVersion.....not needed at all, at least per my testing

calebmshafer (approved):

PR #104, opened 2022-01-12
Add measure-tools from measure-tools-react

aruniverse:


aruniverse:
The only issue ive seen with this new measure tool is when you select a measurement with the selection tool, it still gives you the option to emphasize/isolate and etc which shouldnt be possible (or i dont think it should)

packages/modules/viewer-react/src/hooks/useUiProviders.tsx
```diff
@@ -31,11 +29,14 @@ export function useUiProviders(
         })
       );
     }
-    // if (
-    //   !defaultUiConfig?.contentManipulationTools?.verticalItems?.measureTools
-    // ) {
-    //   defaultProviders.push(new MeasureToolsUiItemsProvider());
-    // }
+    if (
```
kckst8, line 32: When this is "true" it means that it should not be visible. The variable could be named better, but basically all of the ui config booleans remove content only

aruniverse: discussed internally, this logic is correct, i just needed to add an extra check

kckst8 (approved):

PR #105, opened 2022-01-12
Section clear statusbar item

aruniverse:
Not sure why no one had reported this before Travis. Currently, if you make use of the section tools, there is no way to clear/revert your sectioning without refreshing the viewer/pg

https://i.imgur.com/xdYbIY2.gif


tcobbs-bentley (approved):
I don't have any real familiarity with status fields from appui-react, but this looks good.

kckst8 (approved):

PR #106, opened 2022-01-18
trim deps

aruniverse:


packages/modules/desktop-viewer-react/src/types.ts
```diff
@@ -17,3 +17,9 @@ export enum ModelStatus {
   SNAPSHOT,
   COMPARING,
 }
+
+export interface ProgressInfo {
```
calebmshafer, line 21: I was debating what to do with this type in the core. I ended up wrapping ir as an internal API in the NativeHost setup. Would it make it easier if we just exposed it there instead of you having to re-define it here?

aruniverse: That'd def help; i wasnt sure if you had time to backport it to 3.0.0. 
If it goes into 3.1, then wed need to up our min peer dep req, and thatd be a breaking change i think

cc @kckst8

kckst8: We should definitely use the type from core since we use it with a core Api. Otherwise they can get out of sync and cause unnecessary maintenance. We should prefer an extra peer dep now that can be removed later rather than having to add one later

aruniverse: Heres the pr to export that from core: https://github.com/iTwin/itwinjs-core/pull/3069
Should this PR be held until then, or should this be completed now, and then update once the core pkg gets updated & released?

kckst8 (approved):

packages/modules/desktop-viewer-react/src/types.ts
calebmshafer: Yea that sounds like a good plan to me. Raman has been working on stuff in this section so that refactor he's doing may complicate how long it takes to the get core PR in.

PR #109, opened 2022-01-20
update telemetry service to set userinfo; manually introspect jwt token

aruniverse:


packages/modules/viewer-react/src/services/telemetry/TelemetryService.ts
```diff
@@ -21,20 +37,30 @@ class TelemetryService implements TelemetryClient {
...
+          process.env.IMJS_URL_PREFIX ?? ""
+        }api.bentley.com/users/me`;
+        const resp = await fetch(url, {
+          headers: {
+            Authorization: accessToken,
+          },
+        });
+        if (resp.ok) {
+          const { user }: UserInfo = await resp.json();
+          this._appInsights.setAuthenticatedUserContext(
```
aruniverse, line 53: the old api using introspection on the actual jwt token would return ids we could use, here these arent obfuscated. i think itsok to collect these, especially since the user has to grant permission to use this api, but wanted to check with you guys

calebmshafer: I'm not sure if we should be tracking it in this way. Plus it won't have the desired effect if someone is using a access token that's client creds. Not sure how the API would handle that case.

aruniverse: Yea that why i said in the nextversion.md its optional because of the client-cred workflow. id imagine it throws an error in the client-cred workflow, so itd be tracked as un authenticated user

NextVersion.md
```diff
@@ -44,20 +44,25 @@ The following props are now required.
...
+##### Removed
 
 - `openid`
 - `email`
 - `profile`
 - `organization`
 
+##### Optional
+
+- `users:read`
```
kckst8, line 65: maybe make a note that only if you intend to use the applicationInsightsKey, or something along those lines?

aruniverse: dont need to make an extra req for this and dont need the extra scope anymore either

packages/modules/viewer-react/src/services/telemetry/TelemetryService.ts
aruniverse: updated to introspect the token client side, and now the ids are obfuscated a bit more

kckst8 (approved):

PR #110, opened 2022-01-21
re-add unit tests

aruniverse:


packages/modules/viewer-react/src/tests/components/app-ui/widgets/BasicToolWidget.test.tsx
```diff
@@ -197,13 +184,8 @@ describe("BasicToolWidget", () => {
       10,
       CoreTools.selectElementCommand
     );
-    // TODO 3.0
```
kckst8, line 200: we should be able to add these back now that you added measure tools I think?

kckst8 (approved):

packages/modules/viewer-react/src/tests/components/app-ui/widgets/BasicToolWidget.test.tsx
aruniverse: Will add PBI to add tests for ui Providers later

PR #117, opened 2022-01-27
Add tile admin configuration

kckst8:


aruniverse (approved):
This looks fine. I didn't look at IModelApp startup, but I'm guessing the tile admin gets created there? Asking cause for ToolAdmin, an instance needs to be passed in.

kckst8:
> This looks fine. I didn't look at IModelApp startup, but I'm guessing the tile admin gets created there? Asking cause for ToolAdmin, an instance needs to be passed in.

yep. oddly the 2 are different. One instantiates an instance from provided options and the other accepts an instance
image

PR #119, opened 2022-02-02
initial pass to fix stale viewports

bsteinbk:
@kckst8  - I can still take a pass at moving on viewstate stuff from IModelLoader to DefaultContentGroup provider if you would like me to?


packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -323,7 +322,7 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
         allFrontstages = [...frontstages];
       }
 
-      if (viewState) {
+      if (isMounted.current && viewState && viewState.iModel.isOpen) {
```
StefanRetief, line 325: Should also check for viewState.iModel.isBlank and proceed as necessary

kckst8: I think (hope) we can revisit the dependencies here and avoid this....ideally this would never execute for a closed connection

kckst8:
> @kckst8 - I can still take a pass at moving on viewstate stuff from IModelLoader to DefaultContentGroup provider if you would like me to?

Thanks @bsteinbk . I'd like to take a deeper look at these changes. We really need to refactor IModelLoader completely. It's gotten a bit out of hand. This could be the catalyst. I will let you know when I inevitably have questions

PR #124, opened 2022-02-08
Refactoring IModelLoader

kckst8:
What began as looking into the issue of stale viewports (which will be resolved in the next 3.0 core patch) turned into some refactoring. The changes made should help to avoid unwanted side effects and at the least separate concerns a bit more for future refactoring, without breaking any APIs. There is a "breaking" UI change in that the corner item will no longer be visible if there are no additional frontstages or backstage items. This has been requested by multiple users in the past and I can't think of a reason why this would not be a desired change. That said, it is technically new behavior so I'm ok delaying if need be. 

Still need to write/fix tests but I wanted to start getting some comments. 

NOTE: The ViewportAccessWidget was provided by the Sandbox team to help diagnose the stale Viewport bug. I left it in the test app as is to help test model switching in the future. It should be considered test only

### DIFF
--- a/packages/modules/viewer-react/src/components/app-ui/providers/BackstageItemsProvider.ts
+++ b/packages/modules/viewer-react/src/components/app-ui/providers/BackstageItemsProvider.ts
@@ -1,0 +1,74 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import type { UiItemsProvider } from "@itwin/appui-abstract";
+import type {
+  BackstageActionItem,
+  BackstageStageLauncher,
+} from "@itwin/appui-abstract";
+import { BackstageItemUtilities } from "@itwin/appui-abstract";
+import { UiFramework } from "@itwin/appui-react";
+import { IModelApp } from "@itwin/core-frontend";
+
+import type { ViewerBackstageItem } from "../../../types";
+
+export class BackstageItemsProvider implements UiItemsProvider {
+  public readonly id = "BackstageItemsProvider";
+  private _backstageItems: ViewerBackstageItem[];
+
+  constructor(backstageItems: ViewerBackstageItem[]) {
+    this._backstageItems = backstageItems;
+  }
+
+  public provideBackstageItems() {
+    const allBackstageItems: ViewerBackstageItem[] = [];
+    this._backstageItems.forEach((backstageItem) => {
+      // check for label i18n key and translate if needed
+      if (backstageItem.labeli18nKey) {
+        let newItem;
+        if ((backstageItem as BackstageStageLauncher).stageId) {
+          newItem = BackstageItemUtilities.createStageLauncher(
+            (backstageItem as BackstageStageLauncher).stageId,
+            backstageItem.groupPriority,
+            backstageItem.itemPriority,
+            IModelApp.localization.getLocalizedString(
+              backstageItem.labeli18nKey
+            ),
+            backstageItem.subtitle,
+            backstageItem.icon
+          );
+        } else {
+          newItem = BackstageItemUtilities.createActionItem(
+            backstageItem.id,
+            backstageItem.groupPriority,
+            backstageItem.itemPriority,
+            (backstageItem as BackstageActionItem).execute,
+            IModelApp.localization.getLocalizedString(
+              backstageItem.labeli18nKey
+            ),
+            backstageItem.subtitle,
+            backstageItem.icon
+          );
+        }
+        allBackstageItems.push(newItem);
+      } else {
+        allBackstageItems.push(backstageItem);
+      }
+    });
+
+    // add a launcher item for the built-in frontstage if there is an active connection and other backstage items
+    if (allBackstageItems?.length > 0 && UiFramework.getIModelConnection()) {
+      allBackstageItems.unshift({
+        stageId: "DefaultFrontstage",
+        id: "DefaultFrontstage",
+        groupPriority: 100,
+        itemPriority: 10,
+        label: IModelApp.localization.getLocalizedString(
+          "iTwinViewer:backstage.mainFrontstage"
+        ),
+      });
+    }
+    return allBackstageItems;
+  }
+}
--- a/packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
+++ b/packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
@@ -1,0 +1,107 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { StandardContentLayouts } from "@itwin/appui-abstract";
+import {
+  ContentGroup,
+  ContentGroupProvider,
+  IModelViewportControl,
+  UiFramework,
+} from "@itwin/appui-react";
+import type { IModelConnection, ViewState } from "@itwin/core-frontend";
+import { ViewCreator3d } from "@itwin/core-frontend";
+
+import { createBlankViewState } from "../../../services/iModel/ViewCreatorBlank";
+import type {
+  BlankConnectionViewState,
+  ViewerViewCreator3dOptions,
+  ViewerViewportControlOptions,
+} from "../../../types";
+
+/**
+ * Provide a default content group to the default frontstage
+ */
+export class DefaultContentGroupProvider extends ContentGroupProvider {
+  private _viewportOptions: ViewerViewportControlOptions | undefined;
+  private _blankConnectionViewState: BlankConnectionViewState | undefined;
+  private _viewCreatorOptions: ViewerViewCreator3dOptions | undefined;
+
+  constructor(
+    viewportOptions?: ViewerViewportControlOptions,
+    viewCreatorOptions?: ViewerViewCreator3dOptions,
+    blankConnectionViewStateOptions?: BlankConnectionViewState
+  ) {
+    super();
+    this._viewportOptions = viewportOptions;
+    this._blankConnectionViewState = blankConnectionViewStateOptions;
+    this._viewCreatorOptions = viewCreatorOptions;
+  }
+
+  async getViewState(
+    connection: IModelConnection | undefined
+  ): Promise<ViewState | undefined> {
+    if (!connection || (!connection.isBlank && connection.isClosed)) {
+      return;
+    }
+    let view: ViewState | undefined;
+    if (this._viewportOptions?.viewState) {
+      if (typeof this._viewportOptions?.viewState === "function") {
+        view = await this._viewportOptions?.viewState(connection);
+      } else {
+        view = this._viewportOptions?.viewState;
+      }
+    }
+    if (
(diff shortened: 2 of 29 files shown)


aruniverse:
This PR replaces #119 right?

kckst8:
Correct. I'll close that one

StefanRetief:
Testing against the sample showcase and it does end up behaving differently than previously. While looking at some of the multi-viewport samples, which update the `frontstages` prop within the `onIModelConnected` method, I get some unpredictable results. In some cases, I get a view that doesn't show a correct frontstage and other cases, I get a frontstage that is correct, but the viewports are not working correctly. I'll do some more testing by changing how the frontstage providers are created in the showcase, but this might be more of a breaking change than anticipated

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -71,16 +58,17 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
...
+    const { finalFrontstages, noConnectionRequired } = useFrontstages(
+      frontstages,
+      defaultUiConfig,
+      viewportOptions,
+      viewCreatorOptions,
+      blankConnectionViewState
+    );
 
-    useUiProviders(uiProviders, defaultUiConfig);
+    useUiProviders(uiProviders, defaultUiConfig, backstageItems);
```
aruniverse, line 71: do we need the useEffect in L76 for settingError? couldnt we just do that on L165?

aruniverse: you can disregard this, i really need to look at the error-handling pkg and clean it up

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -104,150 +92,78 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
...
-        }
+      ViewerPerformance.addMark("IModelConnectionStarted");
+      void ViewerPerformance.addAndLogMeasure(
+        "IModelConnecting",
+        "ViewerStarting",
+        "IModelConnectionStarted"
+      );
+      let imodelConnection: IModelConnection | undefined;
+      // create a new imodelConnection for the passed project and imodel ids
+      // TODO add the ability to open a BriefcaseConnection for Electron apps
```
aruniverse, line 122: i think you can remove this comment

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -104,150 +92,78 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
...
     };
 
-    const getViewState = useCallback(async () => {
-      if (!connection || (!connection.isBlank && connection.isClosed)) {
-        setViewState(undefined);
-        return;
+    const getModelConnection = useCallback(async (): Promise<
+      IModelConnection | undefined
+    > => {
+      if (blankConnection) {
```
aruniverse, line 101: should you check that its mounted too?

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -259,107 +175,20 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
...
-        // add the default frontstage first so that it's default status can be overridden
-        allFrontstages.unshift({
-          provider: defaultFrontstageProvider,
-          default: true,
-        });
-      }
-
-      setFinalFrontstages(allFrontstages);
-    }, [frontstages, viewportOptions, viewState, defaultUiConfig]);
+    }, [getModelConnection]);
```
aruniverse, line 178: missing dep on isMounted

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -259,107 +175,20 @@ const Loader: React.FC<ModelLoaderProps> = React.memo(
...
 
     if (error) {
       throw error;
     } else {
       return (
         <div className="itwin-viewer-container">
           {finalFrontstages &&
-          finalBackstageItems &&
-          ((viewState && connection) || noConnection) &&
+          (connection || noConnectionRequired) &&
```
aruniverse, line 186: this lines a doozy

aruniverse (commented):
tested it out, and it works fine. Using these new Provider classes was on the to do list for  viewer@3.0 anyway,

kckst8:
> Testing against the sample showcase and it does end up behaving differently than previously. While looking at some of the multi-viewport samples, which update the `frontstages` prop within the `onIModelConnected` method, I get some unpredictable results. In some cases, I get a view that doesn't show a correct frontstage and other cases, I get a frontstage that is correct, but the viewports are not working correctly. I'll do some more testing by changing how the frontstage providers are created in the showcase, but this might be more of a breaking change than anticipated

Thanks for testing @StefanRetief ! Can you send me your repo where I can try to reproduce. Might make it easier to track down the bugs

packages/modules/viewer-react/src/components/app-ui/providers/BackstageItemsProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  constructor(backstageItems: ViewerBackstageItem[]) {
+    this._backstageItems = backstageItems;
+  }
+
+  public provideBackstageItems() {
+    const allBackstageItems: ViewerBackstageItem[] = [];
+    this._backstageItems.forEach((backstageItem) => {
+      // check for label i18n key and translate if needed
+      if (backstageItem.labeli18nKey) {
```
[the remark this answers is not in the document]
kckst8: this accepts an i18n key instead of string (basically when this is used, we localize it).

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+      if (typeof this._viewportOptions?.viewState === "function") {
+        view = await this._viewportOptions?.viewState(connection);
+      } else {
+        view = this._viewportOptions?.viewState;
+      }
+    }
+    if (
+      !this._viewportOptions?.alwaysUseSuppliedViewState &&
+      (!view ||
+        (view.iModel.iModelId !== connection.iModelId && connection.isOpen))
```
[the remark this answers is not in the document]
kckst8: I don't think so. All of this logic should apply to all of those cases.

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
kckst8: this extra function was unnecessary. I removed it

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
kckst8: updated

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
kckst8: fixed

packages/modules/viewer-react/src/components/app-ui/providers/BackstageItemsProvider.ts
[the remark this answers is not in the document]
aruniverse: Yea I'm questioning if we should be the ones to localize it. The label can be a cinditionakstring value or whatever it's called in core. Can ignore for now, something to revisit in 3.0

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Provide a default content group to the default frontstage
+ */
+export class DefaultContentGroupProvider extends ContentGroupProvider {
+  private _viewportOptions: ViewerViewportControlOptions | undefined;
+  private _blankConnectionViewState: BlankConnectionViewState | undefined;
+  private _viewCreatorOptions: ViewerViewCreator3dOptions | undefined;
+
+  constructor(
```
aruniverse, line 29: nit: i think we should start passing objects around instead of multiple args in the future, in next

packages/modules/viewer-react/src/hooks/useFrontstages.tsx
```diff
@@ -0,0 +1,84 @@
...
+          }
+        }
+        if (!requiresConnection) {
+          // allow to continue to render
+          setNoConnectionRequired(true);
+        }
+      }
+    }
+
+    if (!noConnectionRequired) {
```
aruniverse, line 57: does this get update immediately? might as well just pull `requiresConnection` up higher in scope, and check against that

packages/modules/viewer-react/src/services/iModel/IModelService.ts
```diff
@@ -79,3 +88,80 @@ export const openLocalImodel = async (fileName: string) => {
...
+    } else {
+      view = viewportOptions?.viewState;
+    }
+  }
+  if (
+    !viewportOptions?.alwaysUseSuppliedViewState &&
+    (!view ||
+      (view.iModel.iModelId !== connection.iModelId && connection.isOpen))
+  ) {
+    if (connection.isBlankConnection()) {
```
aruniverse, line 148: nit: inconsistent with L132,

aruniverse (approved):
didnt pull to test but code looks fine; @StefanRetief please approve once you confirm this resolves the sandbox issues

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
aruniverse: comment applies i think for a lot of the fns, classes where we have multiple args. but this is a huge nit, and something to tackle later

packages/modules/viewer-react/src/services/iModel/IModelService.ts
kckst8: I'll update but they are actually for different cases. The first case we don't care about type guards so the bool suffices. In the second, we are passing a BlankConnection to the function, so isBlankConnection will provide a type guard for us

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
kckst8: I agree, but just need to be careful that there is no mutation (which there shouldn't be anyways but I have seen it in some code that we work with often :-)) and that the object isn't passed along too many times

aruniverse: I hope not my code lol

StefanRetief (approved):
LGTM, all samples work well against these changes.

packages/modules/viewer-react/src/hooks/useFrontstages.tsx
kckst8: updated. I ended up changing the logic here to be simpler. We actually take a last is first approach to multiple default frontstages (see [IModelViewer.tsx](https://github.com/iTwin/viewer/blob/viewport-issue/packages/modules/viewer-react/src/components/iModel/IModelViewer.tsx#L32)) and here we were taking first is first.

packages/modules/viewer-react/src/services/iModel/IModelService.ts
kckst8: updated

PR #125, opened 2022-02-09
rm default ui providers

aruniverse:


packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
```diff
@@ -80,6 +98,14 @@ export const ViewerHome: React.FC = () => {
           },
         }}
         enablePerformanceMonitors={true}
+        onIModelAppInit={onIModelAppInit}
+        uiProviders={[
+          new TreeWidgetUiItemsProvider(),
+          // new PropertyGridUiItemsProvider({
```
kckst8, line 104: waiting on something here?

packages/apps/web-viewer-test/tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx",
+    "jsx": "react",
```
kckst8, line 20: I think we should leave this (for React 17). Any reason to the contrary?

kckst8 (approved):
few minor questions/comments. Looks good otherwise. Finally, you get your wish :-)

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
aruniverse: https://github.com/iTwin/viewer-components-react/pull/185

Cant init PropertyGrid yet

aruniverse:
> Finally, you get your wish :-)

😁 Let me know if you think this is ok to go into `next` or if we should defer it

PR #126, opened 2022-02-11
Modified onIModelConected to await async method when available

StefanRetief:


aruniverse (commented):
dup headers, need rush change, minor

aruniverse (approved):

kckst8 (approved):

PR #127, opened 2022-02-16
update default imodel loader indicator

aruniverse:


packages/modules/viewer-react/src/components/iModel/IModelLoader.scss
```diff
@@ -14,6 +14,11 @@
   justify-content: center;
 }
 
+.itwin-viewer-loading-icon {
+  width: 64px;
```
aruniverse, line 18: lmk if too small

kckst8: LGTM

kckst8 (approved):

PR #128, opened 2022-02-17
Fix backstage items label bug

kckst8:


packages/modules/viewer-react/src/components/iModel/IModelViewer.tsx
```diff
@@ -58,9 +58,7 @@ export const IModelViewer: React.FC<ModelProps> = ({
         <ConfigurableUiContent
           appBackstage={
             backstageItems &&
-            backstageItems.length > 0 && (
-              <BackstageComposer items={backstageItems} />
-            )
+            backstageItems.length > 0 && <BackstageComposer items={[]} />
```
aruniverse, line 61: could we possibly have used this to hide the backstage button? https://github.com/iTwin/itwinjs-core/blob/21f65735377a4770d362ce912c47047acd10294a/ui/appui-react/src/appui-react/backstage/BackstageComposer.tsx#L129

aruniverse (approved):

packages/modules/viewer-react/src/components/iModel/IModelViewer.tsx
kckst8: this is specific to stage launchers. We take it a step further and hide if there is only one backstage item (launcher or action). Also, looks like this still renders the parent component

PR #130, opened 2022-03-01
Remove unnecessary Webpack plugin sections from all Viewer READMEs

johnnyd710:
Removed the Webpack plugin details in the "Dependencies" section of the *web-viewer-react* and *desktop-viewer-react* READMEs as they were no longer necessary.

### DIFF
--- a/packages/modules/web-viewer-react/README.md
+++ b/packages/modules/web-viewer-react/README.md
@@ -16,7 +16,7 @@ npm install @itwin/web-viewer-react
 
 ## Dependencies
 
-If you are creating a new application and are using React, it is advised that you use create-react-app with @bentley/react-scripts, which already include this plugin, as well as some other optimizations. There is also a predefined template that includes the iTwin Viewer package:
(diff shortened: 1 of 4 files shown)


packages/modules/web-viewer-react/README.md
```diff
@@ -16,16 +16,6 @@ npm install @itwin/web-viewer-react
...
-
-In your webpack.config file:
-
-```javascript
-    plugins: [
-      // NOTE: iTwin.js specific plugin to allow exposing iTwin.js shared libraries into the global scope.
-      new IModeljsLibraryExportsPlugin(),
-```
-
 If you are creating a new application and are using React, it is advised that you use create-react-app with @bentley/react-scripts, which already include this plugin, as well as some other optimizations. There is also a predefined template that includes the iTwin Viewer package:
```
aruniverse, line 19: Looks good, you might want to remove the reference to the plugin in this line tho. Same for desktop.

aruniverse (approved):
looks good!

PR #131, opened 2022-03-02
Do not use the ViewCreator seed by default

johnnyd710:
Removed the code that defaults the useSeedView option to "true", but still allows it as an option. Since we are no longer creating a new object with useSeedView set to true, we can directly pass viewCreatorOptions to the createDefaultView function.


kckst8 (approved):
great work!

PR #148, opened 2022-03-09
use StandardFrontstageProvider from core instead of custom one

aruniverse:
Use [StandardFrontstageProvider](https://github.com/iTwin/itwinjs-core/blob/master/ui%2Fappui-react%2Fsrc%2Fappui-react%2Ffrontstage%2FStandardFrontstageProvider.tsx) from core instead of our default which pre-populates many tools

use [standard ui-providers from core](https://github.com/iTwin/itwinjs-core/tree/master/ui/appui-react/src/appui-react/ui-items-provider) to populate default items in the test -apps for now

Todo: In a separate pr add & export UiProviders to the base viewer to match existing look & behavior of the viewer (core's standard status bar provider doesnt match the viewer's defaults)


kckst8 (commented):
I think this makes sense, but not sure if we want to do it yet. Should we decide what to do about defaultUIConfig first (or maybe wait until after we meet with UX)? I feel like there will still be a case for disabling some items if we continue with pre-canned templates, but maybe that is external configuration driven instead of a prop? My thought was that template responses could drive a config file that lists tools, widgets, etc (eventually replaced by extensions) that the Viewer requests on startup (or is bundled). 

I'm not sure if we want to remove the Blank Viewer component, but maybe we need to discuss. It was one originally and then it made sense to extract it since it has less stringent requirements for those users (authClient, etc) rather than having them parse a bunch of unnecessary docs

aruniverse:
Yea I dont think this pr should complete until after we determine what the default ui should look like, merely opening this draft pr in prep for that call. 

For the pre-canned templates, i was thinking responses would be used as props for the ui providers

Re blank/viewer consolidation, i think we could get away with it with some type magic, havent looked into it yet, but should be feasible i think. Also why i was bringing up the public models. One viewer component that handles all the different cases, all the real work is done by the IModelLoader anyway

packages/modules/viewer-react/src/hooks/useFrontstages.tsx
```diff
@@ -54,13 +57,25 @@ export const useFrontstages = (
 
     if (requiresConnection) {
       // we require a connection, so initialize the DefaultFrontstage that contains the views that we want
-      const defaultFrontstageProvider = new DefaultFrontstage(
-        defaultUiConfig,
+      // const defaultFrontstageProvider = new DefaultFrontstage(
```
kckst8, line 60: can just delete this unless there is reason to keep it around

aruniverse: Thought id leave this around for testing the new ui providers in the todo; was thinking we can remove this chunk once we're at a feature/look parity with whats on master

PR #149, opened 2022-03-09
consolidate Viewer & BlankViewer

aruniverse:


packages/modules/viewer-react/src/components/BaseViewer.tsx
```diff
@@ -55,24 +35,10 @@ export const BaseViewer: React.FC<ViewerProps> = ({
   const accessToken = useAccessToken();
   return (
     <ErrorBoundary>
-      {snapshotPath || accessToken ? (
+      {("filePath" in iModelLoaderProps && iModelLoaderProps.filePath) ||
```
johnnyd710, line 38: Do we need the first part of this check (`"filePath" in iModelLoaderProps`)? What if iModelLoaderProps is undefined for some reason - it would throw an error - although that would be unlikely it seems

packages/modules/web-viewer-react/src/types.ts
```diff
@@ -36,19 +37,12 @@ export interface IModelBackendOptions {
   customBackend?: CustomBackendConfig;
 }
 
-export interface WebViewerPropsFull extends ViewerProps {
+export type WebProps = XOR<ConnectedViewerProps, BlankViewerProps> & {
```
kckst8, line 40: Much better

packages/modules/viewer-react/src/components/BaseViewer.tsx
aruniverse: Yea this was really annoying, and probably the ugliest part of all my changes. I was hoping it could infer `filePath` existed from the union, but it couldn't auto infer it because of the union. Needed to use the `in` keyword to get the [discriminated union](https://www.typescriptlang.org/play?q=267#example/discriminate-types) to work. I need to see if I can clean this up with the `XOR`

packages/apps/web-viewer-test/src/components/home/IModelBankHome.tsx
```diff
@@ -106,8 +106,8 @@ export const IModelBankHome: React.FC = () => {
     <div style={{ height: "100vh" }}>
       <Viewer
         authClient={authClient}
-        iTwinId={iTwinId}
-        iModelId={iModelId}
+        iTwinId={iTwinId ?? ""}
```
kckst8, line 109: why the need for empty strings? Can't be left undefined?

packages/modules/desktop-viewer-react/README.md
```diff
@@ -48,9 +48,9 @@ export const MyViewerComponent = () => {
 #### Optional
 
 - `iTwinId` - GUID for the iTwin (project, asset, etc.) that contains the iModel that you wish to view
-- `iModelId` - GUID for the iModel that you wish to view. Must be specified with a iTwinId if a snapshotPath is not provided
+- `iModelId` - GUID for the iModel that you wish to view. Must be specified with a iTwinId if a filePath is not provided
```
kckst8, line 51:
```suggestion
- `iModelId` - GUID for the iModel that you wish to view. Must be specified with an iTwinId if a filePath is not provided
```

packages/modules/viewer-react/src/types.ts
```diff
@@ -16,19 +16,23 @@ import type {
...
-  TileAdmin,
-  ToolAdmin,
   ViewChangeOptions,
   ViewCreator3dOptions,
   ViewState,
 } from "@itwin/core-frontend";
 import type { Vector3d, XAndY, XYAndZ } from "@itwin/core-geometry";
 
+export type Without<T1, T2> = { [P in Exclude<keyof T1, keyof T2>]?: never };
+// eslint-disable-next-line @typescript-eslint/ban-types
```
kckst8, line 31: What is the linter's issue with this? Is something here poor practice?

packages/modules/web-viewer-react/src/services/ItwinViewer.ts
```diff
@@ -57,7 +57,11 @@ export class ItwinViewer {
     this.authClient = options.authClient;
     this.enablePerformanceMonitors = options.enablePerformanceMonitors;
 
-    void WebInitializer.startWebViewer(options);
+    void WebInitializer.startWebViewer({
+      ...options,
+      iTwinId: "",
```
kckst8, line 62: not following why the empty strings are needed?

packages/modules/web-viewer-react/src/services/ItwinViewer.ts
```diff
@@ -66,23 +70,25 @@ export class ItwinViewer {
...
-        onIModelConnected: this.onIModelConnected,
-        frontstages: this.frontstages,
-        viewportOptions: this.viewportOptions,
-        uiProviders: this.uiProviders,
-        theme: this.theme,
-        enablePerformanceMonitors: this.enablePerformanceMonitors,
-      } as WebViewerProps),
-      document.getElementById(this.elementId)
-    );
+    // ReactDOM.render(
```
kckst8, line 89: shouldnt this be un-commented? We'll need to render something :-)

packages/modules/desktop-viewer-react/src/types.ts
```diff
@@ -3,9 +3,9 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import type { ViewerProps } from "@itwin/viewer-react";
+import type { FileViewerProps } from "@itwin/viewer-react";
 
-export type DesktopViewerProps = Omit<ViewerProps, "appInsightsKey">;
+export type DesktopViewerProps = Omit<FileViewerProps, "appInsightsKey">;
```
kckst8, line 8: the issue here is that currently the desktop viewer supports online viewing and we have product teams who are using that functionality. We have to think long and hard about taking this hard stance of only file support in the desktop. I understand that there are some that don't agree within online viewing from the desktop but is it in the best interest of our users to remove support completely?

kckst8 (commented):
Initially the BlankViewer was added independently due to its large deviation from the current Viewer (even from an ideological standpoint). There has been discussion about making reality data a first class citizen in our services and I made the assumption that it could involve some changes that we don't necessarily want to build into the more standard viewer. That, and giving users an explicit declaration for blank connection scenarios, while also limiting the large list of standard Viewer props which are not applicable for that use case. 

There are pros and cons to smaller focused components and larger multi-responsibility components. I think this case is fine but something to keep in mind for the future. Ideally, we want to think user experience first, and our maintainability second. 

Overall looking like a very nice refactor so far. Nice work!

packages/modules/viewer-react/src/tests/components/BaseViewer.test.tsx
```diff
@@ -164,34 +164,34 @@ describe("BaseViewer", () => {
     );
   });
 
-  it("ensures that either a iTwinId/iModelId combination or a local snapshot is provided", async () => {
-    const events = {
-      onError: (event: ErrorEvent) => {
-        event.preventDefault();
-      },
-    };
+  // it("ensures that either a iTwinId/iModelId combination or a local snapshot is provided", async () => {
```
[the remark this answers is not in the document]
aruniverse: Yea will uncomment before it merges

thanks to the xor it makes some of the args required so it actually kinda satisifes what you were checking. users have to define either itwinId/imodelId or filePath or blankConnection

packages/modules/web-viewer-react/src/services/ItwinViewer.ts
```diff
@@ -17,7 +17,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 
 import { Viewer } from "../components/Viewer";
-import type { ItwinViewerParams, WebViewerProps } from "../types";
+import type { ItwinViewerParams, WebProps } from "../types";
```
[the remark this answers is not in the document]
aruniverse: Good point, will rename this

packages/apps/web-viewer-test/src/components/home/IModelBankHome.tsx
aruniverse: because of the xor, these args are now required, and process.env.VAR is potentially undefined

packages/modules/web-viewer-react/src/services/ItwinViewer.ts
aruniverse: For some reason it was complaining about one of the args being undefined, need to spend more time looking at this;

iirc, this was only used by the "cdn viewer" and i think we need to re-discuss this; was hoping we can remove this, really doesnt feel like this is something we should be delivering

packages/modules/desktop-viewer-react/src/types.ts
aruniverse: I thought with the latest desktop viewer we were always opening models againt the local backend, and never the hosted one. I do think we should support blank connections here. I wouldnt be opposed to connected models here either, but yea agreed, lets go over this in more detail with stakeholders

packages/modules/viewer-react/src/components/BaseViewer.tsx
aruniverse: Looks like there is no other way to get around this than using the `"in" in loaderProps`  due to the union :/ . `loaderProps` should never be undefined due to the xor, it should always have the required props for the specific viewer (connected/blank/file)

packages/modules/viewer-react/src/types.ts
aruniverse: Record worked fine, i dont love using `unknown` but eh. resolving this comment

packages/modules/viewer-react/src/tests/components/BaseViewer.test.tsx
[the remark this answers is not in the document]
aruniverse: Removed test due to it becoming obsolete; previously all props were optional, but now with the xor some are required

packages/modules/desktop-viewer-react/src/types.ts
aruniverse: Updated so desktop can open any kind of model

packages/modules/web-viewer-react/src/services/ItwinViewer.ts
aruniverse: this is the only thing left to do in the pr i think; im in favor of removing this entirely; but we can take the discussion offline

kckst8: this is the CDN viewer. We can't remove it without a suitable replacement

aruniverse: yea, but have we even published an updated cdn viewer in a yr? is anyone actually using the cdn viewer?

kckst8: no and maybe :-). It's still being marketed to potential users though. We'll need to support it still (at least for the short term). It's more on me not being able to prioritize updating the pipelines why it hasnt been deployed

aruniverse: I cant wait to delete the cdn viewer :) I'll try to survive until then.

Updated the class

kckst8 (approved):

Issue #133, opened 2022-03-04
Synchronize tsconfig targets in the monorepo


kckst8:
multiple different targets are being used. All should be "es2020". Ensure that everything builds and runs after the change

aruniverse:
Take a look at: 
- https://kangax.github.io/compat-table/es2016plus/ & sort by `Engine types`
- [What TypeScript configuration produces output closest to Node.js 16 capabilities?](https://stackoverflow.com/questions/67371787/what-typescript-configuration-produces-output-closest-to-node-js-16-capabilities)
   - I would keep the module as `ES2020`, and would not want to set `"type": "module"` in the package.json yet, I don't want to start importing from `*.js` files yet :/

In the tsconfig for `itwin/viewer-react`, `itwin/web-viewer-react, `itwin/desktop-viewer-react`, they extend the one found in [core](https://github.com/iTwin/itwinjs-core/blob/master/tools/build/tsconfig-base.json), so you will need to manually add an entry for `"target": "ES2020"` to override whats in the base.

Please test the web-viewer-test app in the repo, I think last summer when i tried to update to es2020, webpack4 didnt work with it.

PR #150, opened 2022-03-11
add local and remote test extensions

kckst8:
- add local and remote test extensions
- update extension APIs


packages/apps/web-viewer-test/tsconfig.json
```diff
@@ -1,11 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "target": "es2020",
```
aruniverse, line 3: Nice! I was hoping to make this the default in itwin/build-tools/tsconfig-base but babel had issues parsing some of the new code. But should work with webpack 5, and I'd like to see if I can change the default for 3.2 if it doesn't cause a breaking change

aruniverse (commented):
Changes looks fine, will test Monday

packages/apps/web-viewer-test/tsconfig.json
kckst8: this one should be done sooner rather than later as well. The targets in this repo are scattered: https://github.com/iTwin/viewer/issues/133. Maybe we should wait until we decide if we can go 2020 across the board

packages/modules/desktop-viewer-react/src/services/Initializer.ts
```diff
@@ -76,7 +75,7 @@ export class DesktopInitializer {
         // register extensions before startup
         // TODO only build time ATM
         if (options?.extensions) {
-          yield addExtensions(options.extensions);
+          yield IModelApp.extensionAdmin.addExtensions(options.extensions);
```
kckst8, line 78: should we remove this yield?

packages/modules/web-viewer-react/src/services/Initializer.ts
```diff
@@ -96,12 +95,16 @@ export class WebInitializer {
...
         );
-        // register extensions before startup
-        // TODO only build time ATM
+        yield IModelApp.startup(iModelAppOptions);
+        // register extensions after startup
         if (options?.extensions) {
-          yield addExtensions(options.extensions);
+          options.extensions.map((extension) => {
+            if (extension.hostname) {
+              IModelApp.extensionAdmin.registerHost(`http://${extension.hostname}`);
```
kckst8, line 103: do we need http here? could we just use the hostname as is? Most cases would be https anyways

kckst8 (commented):
few minor comments

packages/modules/web-viewer-react/src/services/Initializer.ts
johnnyd710: registerHost() expects a URL so without the protocol it errors. ... but it would have been better if registerHost() was flexible enough to accept both

packages/modules/desktop-viewer-react/src/services/Initializer.ts
johnnyd710: Yes my mistake I forgot the desktop initialization entirely...

packages/modules/web-viewer-react/src/services/Initializer.ts
kckst8: Ok, I created an[ issue](https://github.com/iTwin/itwinjs-backlog/issues/61) to change that. now that you are parsing the url in the provider constructors, we shouldn't need to do it again. This is ok for now

packages/modules/desktop-viewer-react/src/services/Initializer.ts
```diff
@@ -73,12 +72,20 @@ export class DesktopInitializer {
         window.onoffline = () => {
           /* nop */
         };
-        // register extensions before startup
-        // TODO only build time ATM
+        yield ElectronApp.startup(electronViewerOpts);
+        // register extensions after startup
```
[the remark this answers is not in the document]
aruniverse: disregard, i havent looked at viewer code in a long time 😅 😭

PR #152, opened 2022-03-15
create default ui providers for the viewer to match existing look

johnnyd710:
Creating this PR to get some feedback. I still have to delete old widgets, status bar, and frontstages and update the corresponding tests.

### DIFF
--- a/packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
+++ b/packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
@@ -1,0 +1,65 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import type { CommonStatusBarItem } from "@itwin/appui-abstract";
+import { StatusBarSection } from "@itwin/appui-abstract";
+import type { DefaultContentTools } from "@itwin/appui-react";
+import {
+  SectionsStatusField,
+  StandardContentToolsProvider,
+  StatusBarItemUtilities,
+  withStatusFieldProps,
+} from "@itwin/appui-react";
+import * as React from "react";
+
+export class ViewerContentToolsProvider extends StandardContentToolsProvider {
+  private _viewerDefaults: DefaultContentTools;
+
+  constructor(defaultItems?: DefaultContentTools) {
+    let viewerDefaults = defaultItems;
+    // by default, don't show the measure tools
+    if (!viewerDefaults) {
+      viewerDefaults = {
+        horizontal: {
+          clearSelection: true,
+          clearDisplayOverrides: true,
+          hide: "element",
+          isolate: "element",
+          emphasize: "element",
+        },
+        vertical: {
+          selectElement: true,
+          measureGroup: false,
+          sectionGroup: true,
+        },
+      };
+    }
+    super("ViewerDefaultContentTools", viewerDefaults);
+    this._viewerDefaults = viewerDefaults;
+  }
+
+  // need to override this method to move sectioning "clear" tool to its proper position on the left
+  public override provideStatusBarItems(): CommonStatusBarItem[] {
+    const statusBarItems: CommonStatusBarItem[] = [];
+
+    // if the sectionGroup tools are to be shown then we want the status field added to allow clearing or manipulation the section
+    if (
+      !this._viewerDefaults ||
+      !this._viewerDefaults.vertical ||
+      this._viewerDefaults.vertical.sectionGroup
+    ) {
+      const Sections = withStatusFieldProps(SectionsStatusField);
+      statusBarItems.push(
+        StatusBarItemUtilities.createStatusBarItem(
+          "uifw.Sections",
+          StatusBarSection.Left,
+          30,
+          <Sections hideWhenUnused />
+        )
+      );
+    }
+
+    return statusBarItems;
+  }
+}
--- a/packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
+++ b/packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
@@ -1,0 +1,133 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import type {
+  CommonStatusBarItem,
+  UiItemsProvider,
+} from "@itwin/appui-abstract";
+import { StatusBarSection } from "@itwin/appui-abstract";
+import { FooterSeparator } from "@itwin/appui-layout-react";
+import type { DefaultStatusbarItems, StatusBarItem } from "@itwin/appui-react";
+import {
+  FooterModeField,
+  MessageCenterField,
+  SelectionInfoField,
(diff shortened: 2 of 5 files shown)


packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+import type { DefaultContentTools } from "@itwin/appui-react";
+import {
+  SectionsStatusField,
+  StandardContentToolsProvider,
+  StatusBarItemUtilities,
+  withStatusFieldProps,
+} from "@itwin/appui-react";
+import * as React from "react";
+
+export class ViewerContentToolsProvider extends StandardContentToolsProvider {
```
aruniverse, line 16: I need to rethink subclassing BaseUiItemsProvider

packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+    // if the sectionGroup tools are to be shown then we want the status field added to allow clearing or manipulation the section
+    if (
+      !this._viewerDefaults ||
+      !this._viewerDefaults.vertical ||
+      this._viewerDefaults.vertical.sectionGroup
+    ) {
+      const Sections = withStatusFieldProps(SectionsStatusField);
+      statusBarItems.push(
+        StatusBarItemUtilities.createStatusBarItem(
+          "uifw.Sections",
```
aruniverse, line 55: whats this arg? i dont think we want to expose "uifw" (ui-framework)

packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
```diff
@@ -0,0 +1,133 @@
...
+  withStatusFieldProps,
+} from "@itwin/appui-react";
+import * as React from "react";
+
+export class ViewerStatusbarItemsProvider implements UiItemsProvider {
+  public readonly id = "ViewerDefaultStatusbar";
+  private _footerModeOnlySeparator: () => React.ReactNode;
+
+  constructor(private _defaultItems?: DefaultStatusbarItems) {
+    const FooterOnlyDisplay = withStatusFieldProps(FooterModeField);
```
aruniverse, line 31: what is this doing?

packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
```diff
@@ -0,0 +1,133 @@
...
+      return (
+        <FooterOnlyDisplay>
+          <FooterSeparator />
+        </FooterOnlyDisplay>
+      );
+    };
+  }
+
+  public provideStatusBarItems(): CommonStatusBarItem[] {
+    const ToolAssistance = withStatusFieldProps(ToolAssistanceField);
```
aruniverse, line 42: nit: id probably only create these if theyre being used (inside the if statements)

packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const MessageCenter = withMessageCenterFieldProps(MessageCenterField);
+    const TileLoadIndicator = withStatusFieldProps(TileLoadingIndicator);
+    const SelectionScope = withStatusFieldProps(SelectionScopeField);
+    const SnapMode = withStatusFieldProps(SnapModeField);
+    const SelectionInfo = withStatusFieldProps(SelectionInfoField);
+    const items: StatusBarItem[] = [];
+
+    if (!this._defaultItems || this._defaultItems.messageCenter) {
```
aruniverse, line 50: could probably have on if statement wrapping all the others to check for _defaultItems

packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
johnnyd710: Originally I had `ViewerContentToolsProvider` as a Ui Provider, but the code was almost the exact same as `StandardContentToolsProvider` (with the exception of the `provideStatusBarItems` function). So I thought extending it and just overriding whatever was needed would result in the least amount of duplicated code.

packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
johnnyd710: Yeah that was just left over from ui-framework, I will remove it

packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
johnnyd710: I think its there so that the vertical separator lines in between the status bar components only show up when the status bar is in "footer mode". Its from the StandardStatusbar. If we assume our status bar will always be in footer mode then I can remove it?

packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
johnnyd710: Since its an `||` and not `&&`, I think that would change the behaviour.

kckst8: based on our conversation this morning, can we just remove the message center?

johnnyd710: I think that was about the `Activity Center`, not the `Message Center`. I already removed the `Activity Center` so you won't see it in this PR

packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <FooterSeparator />
+        </FooterOnlyDisplay>
+      );
+    };
+  }
+
+  public provideStatusBarItems(): CommonStatusBarItem[] {
+    const items: StatusBarItem[] = [];
+
+    if (!this._defaultItems || this._defaultItems.messageCenter) {
```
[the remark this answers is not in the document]
johnnyd710: I think that would change the functionality we want? _if defaultItems is undefined or message center was explicitly set, we want the messageCenter to show_:
```
if (!this._defaultItems || this._defaultItems.messageCenter) console.log('true')
```
- prints 'true' if defaultItems is: `undefined`, or `{messageCenter: true}`
- doesn't print anything if defaultItems is `{}`, `{ messageCenter: false }`, etc.  
```
if (this._defaultItems?.messageCenter) console.log('true')
```
- prints 'true' __only__ if defaultItems is `{messageCenter: true}`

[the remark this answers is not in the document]
kckst8: you're right. Misread

kckst8 (approved):

packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
aruniverse: Yea for now this is fine, and i think probably what we'll end up doing.

The BaseUiItemsProvider class has an extra arg for chekcing frontstage support, and that opens up a can of worms...

But for now, this is fine, and lgtm

PR #160, opened 2022-04-18
Remove telemetry and dependency on Microsoft's Application Insights

johnnyd710:
Saves ~130 Kb in the final bundle size.


packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -207,23 +206,8 @@ const Loader = React.memo(
...
-  appInsightsKey?: string;
-};
-
-const IModelLoader = (props: IModelLoaderProps) => {
-  if (props.appInsightsKey) {
-    return <TrackedLoader {...props} />;
-  } else {
-    return <Loader {...props} />;
-  }
+const IModelLoader = (props: ModelLoaderProps) => {
```
aruniverse, line 209: we dont need this wrapped component anymore, can just use Loader, but i guess you can rename it to IModelLoader.

@kckst8, preference on defaults vs named export?

kckst8: since this file has a single export with no real potential for others and its "internal", I typically prefer default. That said, its not a strong preference and I wouldn't argue named exports

packages/modules/viewer-react/src/tests/components/BaseViewer.test.tsx
```diff
@@ -34,16 +34,6 @@ jest.mock("@itwin/presentation-frontend", () => {
 });
 jest.mock("react-i18next");
 
-jest.mock("@microsoft/applicationinsights-react-js", () => ({
```
kckst8, line 37: so nice to be able to get rid of this gross mock everywhere

aruniverse (approved):

PR #129, opened 2022-02-24
Add extensions back

kckst8:

PR #165, opened 2022-05-19
tsconfig target es2020

samejima-san:
- init next branch
- Add extensions back (#129)
- rm default ui providers (#125)
- consolidate Viewer & BlankViewer & types (#149)
- use StandardFrontstageProvider from core instead of custom one (#148)
- create default ui providers for the viewer to match existing look (#152)
- don't publish error handling package
- make the package private
- Applying package updates.
- remove version policy for error handling package
- Applying package updates.
- remove version policy from test-extension
- Applying package updates.
- update error-handling package version
- Applying package updates.
- Remove telemetry and dependency on Microsoft's Application Insights (#160)
- Applying package updates.
- Code split the test apps (#161)
- add local and remote test extensions (#150)
- syncronized tsconfig target to es2020

### DIFF
--- a/packages/apps/cdn-viewer/tsconfig.json
+++ b/packages/apps/cdn-viewer/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2017",
+    "target": "es2020",
     "lib": [
       "dom",
       "dom.iterable",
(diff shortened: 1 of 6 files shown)


kckst8 (commented):
there are a couple more files that have additional parts to their names (tsconfig.development...). Can you update those as well? Should be able to just do a global search for "target"

PR #168, opened 2022-05-23
Add renderSys option for IModelApp

achrysaetos:
renderSys is initialized as undefined.

### DIFF
--- a/packages/modules/viewer-react/src/types.ts
+++ b/packages/modules/viewer-react/src/types.ts
@@ -2,6 +2,10 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+/*---------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


packages/modules/viewer-react/src/types.ts
```diff
@@ -2,6 +2,10 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
```
achrysaetos, line 8: I think pre-commit added this. Should I remove the duplicate?

aruniverse: yes please do

achrysaetos: Done.

aruniverse (approved):

PR #170, opened 2022-05-23
better handle initialization errors

samejima-san:
Changes to better handle initialization errors in situations where the IModelApp is initialized prior to the Viewers being initialized.

### DIFF
--- a/packages/modules/web-viewer-react/src/services/Initializer.ts
+++ b/packages/modules/web-viewer-react/src/services/Initializer.ts
@@ -133,6 +133,8 @@ export class WebInitializer {
           WebInitializer._initializing = false;
           WebInitializer._cancel = undefined;
         });
+    } else if (IModelApp.initialized && !this._initialized) {
+      throw new Error("IModel app initialized before web viewer");
     }
   }
 }
--- a/packages/modules/web-viewer-react/src/services/Initializer.ts
+++ b/packages/modules/web-viewer-react/src/services/Initializer.ts
@@ -133,8 +133,10 @@ export class WebInitializer {
           WebInitializer._initializing = false;
           WebInitializer._cancel = undefined;
         });
-    } else if (IModelApp.initialized && !this._initialized) {
-      throw new Error("IModel app initialized before web viewer");
+    } else if (this._initializing) {
+      return this._initialized;
+    } else {
+      throw new Error("Use the useWebViewInitializer");
     }
   }
 }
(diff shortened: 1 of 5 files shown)


aruniverse (changes requested):
Was this supposed to go into master or next?

samejima-san:
> 

next

aruniverse:
@samejima-san can you update the pr description?

packages/modules/web-viewer-react/src/services/Initializer.ts
```diff
@@ -133,6 +133,8 @@ export class WebInitializer {
           WebInitializer._initializing = false;
           WebInitializer._cancel = undefined;
         });
+    } else if (IModelApp.initialized && !this._initialized) {
```
kckst8, line 136: the cases here should be 

`else if (this._initializing) {
   return this._initialized;
}
else {
// throw the error here.....the error message should notify the user to use the useWebViewInitializer hook in their app instead of calling IModelApp.startup
}
}`

kckst8 (changes requested):
Added a comment. Also, should make similar changes in the Initializer.ts file in the desktop-viewer-react package. It already handles the initializing condition I believe, but it does not handle the condition where IModelApp is already initialized via ElectronApp.startup

packages/modules/web-viewer-react/src/tests/services/Initializer.test.ts
```diff
@@ -126,4 +126,14 @@ describe("Initializer", () => {
...
   });
+
+  it("IModel initialized before Webviewer initializer", async () => {
+    await IModelApp.startup();
+    await WebInitializer.startWebViewer({
+      authClient: new MockAuthorizationClient(),
+      enablePerformanceMonitors: false,
+    });
+    await WebInitializer.initialized;
+    expect("IModel app initialized before web viewer");
```
kckst8, line 137: should expect toThrow to ensure that an error is thrown with the proper message https://jestjs.io/docs/expect

packages/modules/web-viewer-react/src/tests/services/Initializer.test.ts
```diff
@@ -126,4 +126,14 @@ describe("Initializer", () => {
     await WebInitializer.initialized;
     expect(IModelApp.startup).toHaveBeenCalled();
   });
+
+  it("IModel initialized before Webviewer initializer", async () => {
```
kckst8, line 130: maybe add a more descriptive description...think of it as a sentence that starts with "it" (i.e. it("should throw an error when..."))

PR #171, opened 2022-05-24
Revert es2020 bump

aruniverse:
This reverts commit 3e35185.
This reverts commit 88e7e63


aruniverse:
not needed anymore after #173

Issue #135, opened 2022-03-04
Use iTwinUI error components


kckst8:
Research if we can replace usage of the itwin-error-handling-react package in the Viewer. Keep in mind that this package is being used in other services so it cannot be deprecated until usage is removed:

https://itwin.github.io/iTwinUI-react/?path=/story/core-errorpage--bad-gateway

aruniverse:
1. Should be able to remove dep on svgr
2. Nuke the svgs and generated components in packages/modules/error-handling-react/src/images/. All the big error svgs should be found in https://www.npmjs.com/package/@itwin/itwinui-illustrations-react, heres the repo for it: https://github.com/iTwin/iTwinUI-icons/tree/main/packages/illustrations/illustrations
3. In packages/modules/error-handling-react/src/containers/Fallback.tsx, use svgs from itwinui-illustrations-react
4. The above will probably need to refactor packages/modules/error-handling-react/src/components/fallback/

PR #173, opened 2022-05-26
Replace usage of the error-handling-react package

achrysaetos:
Removing imports of ErrorBoundary and useErrorManager from the error-handing-react package in viewer-react. Issue https://github.com/iTwin/viewer/issues/135.

### DIFF
--- a/packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
+++ b/packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
@@ -11,6 +11,11 @@ interface Props {
   eventTitle?: string;
 }
 
+interface Error {
+  code?: string;
+  message?: string;
+}
+
 export class ErrorBoundary extends Component<
   Props,
   {
@@ -22,7 +27,7 @@ export class ErrorBoundary extends Component<
     super(props);
     this.state = {
       fallback: false,
-      error: new Error(),
+      error: {},
     };
   }
 
@@ -38,7 +43,7 @@ export class ErrorBoundary extends Component<
 
   override render(): JSX.Element {
     if (this.state.fallback) {
-      let errorType: ErrorPageType = "generic";
+      const errorType: ErrorPageType = "generic";
       let errorMessage: ErrorPageProps["errorMessage"] = (
         <>
           We can't find the iModel that you are looking for or it does not
@@ -47,80 +52,10 @@ export class ErrorBoundary extends Component<
       );
 
       if (this.state.error.message) {
-        switch (this.state.error.message) {
-          case "error401":
-            errorType = "401";
-            errorMessage = (
-              <>
-                You do not have permission to access this server.
-                <br />
-                Unable to fulfill request.
-              </>
-            );
-            break;
-          case "error403":
-          case "ErrorNoEntitlement":
-            errorType = "403";
-            errorMessage = (
-              <>
-                You do not have permission to access this server.
-                <br />
-                Unable to fulfill request.
-              </>
-            );
-            break;
-          case "error404":
-            errorType = "404";
-            errorMessage = (
-              <>
-                We can not find the iModel that you are looking for or it does
-                not exist.
-                <br />
-                Visit the iModelHub or contact our support team.
-              </>
-            );
-            break;
-          case "error500":
-            errorType = "500";
-            errorMessage = (
-              <>
-                Please retry again. If this continues to happen, please contact
-                our support team or visit the iModelHub.
-              </>
-            );
-            break;
-          case "error503":
-          case "ConnectFailed":
-            errorType = "503";
-            errorMessage = (
-              <>
-                This service is being worked on. Please come back in a little
-                bit or visit iModelHub.
-              </>
-            );
-            break;
-          case "MaintenanceMode":
-            errorType = "503";
-            errorMessage = (
-              <>
-                This service is being worked on. Please come back in a little
-                bit or visit iModelHub.
-              </>
-            );
-            break;
-          default:
-            errorType = "generic";
-            errorMessage = (
-              <>
-                We can't find the iModel that you are looking for or it does not
-                exist. Visit the iModelHub or contact our support team.
-              </>
-            );
-            break;
-        }
+        errorMessage = this.state.error.message;
       }
 
-      return <ErrorPage errorMessage={errorMessage} errorType={errorType} />;
+      return <ErrorPage errorType={errorType} errorMessage={errorMessage} />;
     } else {
       return <>{this.props.children}</>;
     }
--- a/packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
+++ b/packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { ErrorPage } from "@itwin/itwinui-react";
-import type { PropsWithChildren, ReactNode } from "react";
+import type { PropsWithChildren } from "react";
 import React, { Component } from "react";
 
 interface State {
@@ -13,7 +13,7 @@ interface State {
 }
 
 export class ErrorBoundary extends Component<
-  PropsWithChildren<ReactNode>,
+  PropsWithChildren<Record<string, unknown>>,
   State
 > {
   override state: State = {
--- a/packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
+++ b/packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
@@ -2,9 +2,10 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+
 import type { ErrorPageProps, ErrorPageType } from "@itwin/itwinui-react";
 import { ErrorPage } from "@itwin/itwinui-react";
-import React, { Component, Fragment } from "react";
+import React, { Component } from "react";
 
 interface Props {
   eventTitle?: string;
@@ -41,7 +42,7 @@ export class ErrorBoundary extends Component<
       let errorMessage: ErrorPageProps["errorMessage"] = (
         <>
           We can't find the iModel that you are looking for or it does not
-          exist. Visit the iModel HUB or contact our support team.
+          exist. Visit the iModelHub or contact our support team.
         </>
       );
 
@@ -75,7 +76,7 @@ export class ErrorBoundary extends Component<
                 We can not find the iModel that you are looking for or it does
                 not exist.
                 <br />
-                Visit the iModel HUB or contact our support team.
+                Visit the iModelHub or contact our support team.
               </>
             );
             break;
@@ -84,7 +85,7 @@ export class ErrorBoundary extends Component<
             errorMessage = (
               <>
                 Please retry again. If this continues to happen, please contact
-                our support team or visit the iModel HUB.
+                our support team or visit the iModelHub.
               </>
             );
             break;
@@ -94,7 +95,7 @@ export class ErrorBoundary extends Component<
             errorMessage = (
               <>
                 This service is being worked on. Please come back in a little
-                bit or visit iModel HUB.
+                bit or visit iModelHub.
               </>
             );
             break;
@@ -103,7 +104,7 @@ export class ErrorBoundary extends Component<
             errorMessage = (
               <>
                 This service is being worked on. Please come back in a little
-                bit or visit iModel HUB.
+                bit or visit iModelHub.
               </>
             );
             break;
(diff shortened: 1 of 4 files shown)


packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -2,14 +2,17 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
```
achrysaetos, line 8: Will remove these duplicate headers.

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,140 @@
...
+        }
+      }
+
+      return (
+        <ErrorPage
+          errorMessage={errorMessage}
+          errorType={errorType}
+          primaryButtonHandle={() => {
+            console.log("Close");
+          }}
```
achrysaetos, line 128: Not sure how to implement these buttons yet... maybe a redirect onClick?

aruniverse: i dont think you need to add any buttons. arent those optional in that component?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,140 @@
...
+          primaryButtonLabel="Close"
+          secondaryButtonHandle={() => {
+            console.log("Cancel");
+          }}
+          secondaryButtonLabel="Cancel"
+        />
+      );
+    } else {
+      return <Fragment>{this.props.children}</Fragment>;
```
aruniverse, line 137: <> is the same thing as <Fragment>. id prefer using `<>`

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,140 @@
...
+                We can't find the iModel that you are looking for or it does not
+                exist. Visit the iModel HUB or contact our support team.
+              </>
+            );
+            break;
+        }
+      }
+
+      return (
+        <ErrorPage
```
aruniverse, line 123: This is the src for the ErrorPage component: https://github.com/iTwin/iTwinUI-react/blob/main/packages/iTwinUI-react/src/core/ErrorPage/ErrorPage.tsx

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,140 @@
...
+            );
+            break;
+          case "error404":
+            errorType = "404";
+            errorMessage = (
+              <>
+                We can not find the iModel that you are looking for or it does
+                not exist.
+                <br />
+                Visit the iModel HUB or contact our support team.
```
johnnyd710, line 78: iModelHub should be one word

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,140 @@
...
+    if (this.state.fallback) {
+      let errorType: ErrorPageType = "generic";
+      let errorMessage: ErrorPageProps["errorMessage"] = (
+        <>
+          We can't find the iModel that you are looking for or it does not
+          exist. Visit the iModel HUB or contact our support team.
+        </>
+      );
+
+      if (this.state.error.message) {
```
aruniverse, line 48: what is the type of error.message?
i question the error messages here, i guess these are copied from the current fallback we use, but these were pulled out of design review. i wonder if we should just use the message from the error itself.

@kckst8  thoughts?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
achrysaetos: Yep, they are, I'll remove them.

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
aruniverse: @kckst8 , would you be ok if we always used the generic error page: https://itwin.github.io/iTwinUI-react/?path=/story/core-errorpage--generic and the message would just be different and populated by the actual error message?

The current errors thrown don't have http status code unfrotunately, and Id rather not we go through all the possible erros here and map them to the right status code. https://github.com/iTwin/imodels-clients/blob/main/clients/imodels-client-management/src/base/interfaces/IModelsErrorInterfaces.ts

I'll need to talk to Austeja because i dont like they don't report the status code, and they just report their own enum. Feel like all our clients should've standardized error handling

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { ErrorPageProps, ErrorPageType } from "@itwin/itwinui-react";
+import { ErrorPage } from "@itwin/itwinui-react";
+import React, { Component } from "react";
+
+interface Props {
+  eventTitle?: string;
```
aruniverse, line 11: where are we using this?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+interface Error {
```
aruniverse, line 14: do we need this type? isnt this built into typescript?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,63 @@
...
+          We can't find the iModel that you are looking for or it does not
+          exist. Visit the iModelHub or contact our support team.
+        </>
+      );
+
+      if (this.state.error.message) {
+        errorMessage = this.state.error.message;
+      }
+
+      return <ErrorPage errorType={errorType} errorMessage={errorMessage} />;
```
aruniverse, line 58: errortype should always be the generic one right?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return {
+      fallback: true,
+      error: e,
+    };
+  }
+
+  override render(): JSX.Element {
+    if (this.state.fallback) {
+      const errorType: ErrorPageType = "generic";
+      let errorMessage: ErrorPageProps["errorMessage"] = (
```
aruniverse, line 47: what is the type of this supposed to be? is this supposed to be a string or a JSX.Element?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
achrysaetos: So this is what an error page would look like with a generic title and svg, but with the error message populated by the actual error message: 
image
Ideally, we'd want to have a custom title and svg that maps to each unique error code though.

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
achrysaetos: Currently, yes.

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
achrysaetos: It is, but TypeScript's Error interface doesn't have a field for the error code. Let me see if it still works if I revert it.

aruniverse: but are you even using error code?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
aruniverse: ok, take a look at line 55, not sure how youre assigning a string to a react node

aruniverse: disregard, 
image

achrysaetos: That makes sense. When I did console.log(typeof errorMessage) it turned from an object to a string, but no errors. For consistency, I suppose I can put this.state.error.message into a React fragment though.

aruniverse: I guess its fine as is. I question that default message you provide. Error.message should always be defined if an Error exists right?

achrysaetos: Hmm, so we don't need a default message?

aruniverse: well if error.message is always defined, then we'd never have to fall back to the default message right?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,46 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ErrorPage } from "@itwin/itwinui-react";
+import React, { Component } from "react";
+
+interface Props {
```
aruniverse, line 9: looking really good now, i think this might be my last comment, you should be able to remove this interface, and then in line 18, you can use PropsWithChildren instead of Props

achrysaetos: As in PropsWithChildren<ReactNode>?

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,43 @@
...
+import type { PropsWithChildren, ReactNode } from "react";
+import React, { Component } from "react";
+
+interface State {
+  fallback: boolean;
+  error: Error;
+}
+
+export class ErrorBoundary extends Component<
+  PropsWithChildren<ReactNode>,
```
aruniverse, line 16: I believe this should just be `PropsWithChildren<{}>`

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,43 @@
...
+  State
+> {
+  override state: State = {
+    fallback: false,
+    error: new Error(),
+  };
+
+  static getDerivedStateFromError(e: Error): State {
```
aruniverse, line 24: nice job using the state from above instead of redefining it inline :)

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
aruniverse: https://github.com/iTwin/viewer/pull/173#discussion_r886111470

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
achrysaetos: Oh, whenever I use empty brackets pre-commit flags it:
15:64  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead  @typescript-eslint/ban-types

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
achrysaetos: Thanks!

packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx
aruniverse: I think we want to use `Record<string, never>` here

achrysaetos: I get different errors:
Overload 1 of 2, '(props: PropsWithChildren<Record<string, never>> | Readonly<PropsWithChildren<Record<string, never>>>): ErrorBoundary', gave the following error.       
    Type '{ children: Element; }' is not assignable to type 'Readonly<PropsWithChildren<Record<string, never>>>'.
      Property 'children' is incompatible with index signature.
        Type 'Element' is not assignable to type 'never'.

I'll try googling up the issue.

aruniverse: hmm thats strange, try unknown instead of never then

achrysaetos: That works.

PR #175, opened 2022-05-31
force git to normalize line endings to LF on checkin

johnnyd710:
Since we are no longer enforcing LF style line endings (in .pretterrc), we need to force git to normalize all line endings to LF on check in. That way, the repo is normalized and our pre-commit hook to add the copyright header isn't duplicated every time a Windows user makes a commit.


kckst8 (approved):

johnnyd710:
@aruniverse Found a problem with the approach: since line endings are LF after commit, and set to auto before commit... the pre-commit copyright hook still won't work on Windows. 
```* text=auto eol=lf``` should work though, but now we are back to enforcing LF always. But I don't see a problem with that, and at least now no user git configuration is required since its in the .gitattributes file

aruniverse:
gah thats kinda nnoying, but youre right now that we're always enforcing lf and users dont have to do any setup on their end i think itll be a win

aruniverse (approved):
your pr might be blocked with an audit failure, you can update the audit.js file with the cve and you should be good to go!

PR #179, opened 2022-06-02
Raplemie/additional auth description

raplemie:
Adding scopes information so a user that already have a client registered knows what is required without having to analyse the different .env files provided.

Also added a missing `enablePerformanceMonitors` property in the Viewer react example (This parameter is required but was not in the demo code)

### DIFF
--- a/packages/modules/cra-template-desktop-viewer/template/.env
+++ b/packages/modules/cra-template-desktop-viewer/template/.env
@@ -1,5 +1,5 @@
 # ---- Configuration ----
-ITWIN_VIEWER_SCOPE ="imodelaccess:read offline_access imodels:read projects:read realitydata:read"
+ITWIN_VIEWER_SCOPE ="imodelaccess:read imodels:read projects:read realitydata:read"
 ITWIN_VIEWER_CLIENT_ID=""
 ITWIN_VIEWER_REDIRECT_URI=""
 ITWIN_VIEWER_ISSUER_URL="https://ims.bentley.com"
--- a/packages/modules/cra-template-desktop-viewer/template/README.md
+++ b/packages/modules/cra-template-desktop-viewer/template/README.md
@@ -12,15 +12,12 @@ ITWIN_VIEWER_CLIENT_ID="native-xxxxxxxx"
 
 - You should generate a [client](https://developer.bentley.com/register/) to get started. The client that you generate should be for a desktop app and use the following list of apis. You can add the default redirect uri (http://localhost:3000/signin-callback).
 
-- Scopes expected by the viewer are the following:
+- Scopes expected by the viewer are:
 
-  - `imodelaccess:read` is part of Visualization.
-
-  - `imodels:read` is part of iModels API, within Digital Twin Management.
-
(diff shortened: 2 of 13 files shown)


aruniverse (commented):
Could you also make the same changes in the desktop readmes as well?

packages/modules/cra-template-desktop-viewer/template/.env
```diff
@@ -1,5 +1,5 @@
 # ---- Configuration ----
-ITWIN_VIEWER_SCOPE ="imodelaccess:read offline_access imodels:read projects:read realitydata:read"
+ITWIN_VIEWER_SCOPE ="imodelaccess:read imodels:read projects:read realitydata:read"
```
raplemie, line 2: Removed the "offline_access" scope from the environment as when trying to describe how to add it, we realized that this was a ghost scope that is added automagically to native client, but is never listed in the dev portal. The scope is automatically added to the list of required scopes by the client https://github.com/iTwin/auth-clients/blob/main/packages/electron/src/main/Client.ts#L93 so it is now completely hidden to reduce confusion.

kckst8: great find!

packages/modules/cra-template-web-viewer/template/README.md
```diff
@@ -16,6 +16,16 @@ IMJS_AUTH_CLIENT_SCOPES =""
 
 - You can generate a [test client](https://developer.bentley.com/tutorials/web-application-quick-start/#2-register-an-application) to get started.
 
+- Scopes expected by the viewer are `imodelaccess:read`, `imodels:read` and `realitydata:read`.
```
kckst8, line 19: same formatting as above

packages/modules/cra-template-desktop-viewer/template/README.md
```diff
@@ -10,7 +10,17 @@ Prior to running the app, you will need to add OIDC client configuration to the
 ITWIN_VIEWER_CLIENT_ID="native-xxxxxxxx"
 ```
 
-- You should generate a [client](https://developer.bentley.com/register/) to get started. The client that you generate should be for a desktop app and use the Visualization apis. You can add the default redirect uri (http://localhost:3000/signin-callback).
+- You should generate a [client](https://developer.bentley.com/register/) to get started. The client that you generate should be for a desktop app and use the following list of apis. You can add the default redirect uri (http://localhost:3000/signin-callback).
+
+- Scopes expected by the viewer are the following:
+
+  - `imodelaccess:read` is part of Visualization.
```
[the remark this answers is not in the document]
raplemie: I can change it, but you will still have to find iModels, I can remove this level of detail if its too much.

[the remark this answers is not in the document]
raplemie: Changed, did you want the "Enable the following scopes": text too ? (I didnt change it because in this case its not a walkthrough, merely information on what is expected)

kckst8 (approved):

Issue #178, opened 2022-06-02
[next] viewer providers config


aruniverse:
should not need to set the others to true, this is a bug in the default Providers exported by the viewer

_Originally posted by @aruniverse in https://github.com/iTwin/viewer/pull/177#discussion_r888177198_

PR #182, opened 2022-06-02
Default uiProvider options to true if omitted

achrysaetos:
Issue https://github.com/iTwin/viewer/issues/178. Affects ViewerContentToolsProvider and ViewerNavigationToolsProvider.

### DIFF
--- a/packages/apps/desktop-viewer-test/src/frontend/components/routes/ViewerRoute.tsx
+++ b/packages/apps/desktop-viewer-test/src/frontend/components/routes/ViewerRoute.tsx
@@ -40,7 +40,11 @@ export const ViewerRoute = ({ location }: ViewerRouteProps) => {
       filePath={filePath}
       uiProviders={[
         new ViewerNavigationToolsProvider(),
-        new ViewerContentToolsProvider(),
+        new ViewerContentToolsProvider({
+          vertical: {
+            measureGroup: false,
+          },
+        }),
         new ViewerStatusbarItemsProvider(),
         new TreeWidgetUiItemsProvider(),
         new PropertyGridUiItemsProvider({
--- a/packages/modules/cra-template-desktop-viewer/template/src/frontend/components/routes/ViewerRoute.tsx
+++ b/packages/modules/cra-template-desktop-viewer/template/src/frontend/components/routes/ViewerRoute.tsx
@@ -40,7 +40,11 @@ export const ViewerRoute = ({ location }: ViewerRouteProps) => {
       filePath={filePath}
       uiProviders={[
         new ViewerNavigationToolsProvider(),
-        new ViewerContentToolsProvider(),
+        new ViewerContentToolsProvider({
+          vertical: {
+            measureGroup: false,
+          },
+        }),
         new ViewerStatusbarItemsProvider(),
         new TreeWidgetUiItemsProvider(),
         new PropertyGridUiItemsProvider({


packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
```diff
@@ -16,7 +16,16 @@ import * as React from "react";
 
 export class ViewerContentToolsProvider extends StandardContentToolsProvider {
   constructor(private _defaultItems?: DefaultContentTools) {
-    super("ViewerDefaultContentTools", _defaultItems);
+    super(
```
aruniverse, line 19: gah forgot these extended the Providers from core
imo its probably more of a bug on the core side, 


for content tools youre missing defaults for the horizontal tools as well

i think a cleaner way to do something like this:
image

aruniverse:
Unit tests for this would be nice

packages/modules/viewer-react/src/components/app-ui/providers/ViewerContentToolsProvider.tsx
achrysaetos: I think I tried setting them to true initially, but that doesn't work because it overrides any false values passed in through the viewer component:
image
Adding the rest operator also causes the bug for some reason, so I had to remove it.

aruniverse: Well you want to spread after you set the default, I think im spreading at the wrong level in the above screenshot

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
```diff
@@ -129,7 +129,11 @@ const ViewerHome: React.FC = () => {
         onIModelAppInit={onIModelAppInit}
         uiProviders={[
           new ViewerNavigationToolsProvider(),
-          new ViewerContentToolsProvider(),
+          new ViewerContentToolsProvider({
+            vertical: {
+              measureGroup: false,
```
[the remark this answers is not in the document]
achrysaetos: Like this?
image
Then I'd have to import it from @itwin/web-viewer-react.
Does the desktop version use the same tools from the web viewer? How do I test the desktop version to see if it runs?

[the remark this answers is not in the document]
aruniverse: why do you need to import the web-viewer-react? the providers are in viewer-react? isnt that where youre importing from in the web test app and template?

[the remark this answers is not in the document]
achrysaetos: > why do you need to import the web-viewer-react? the providers are in viewer-react? isnt that where youre importing from in the web test app and template?

Oh we currently have them imported from the package in both the web test app and the template:
image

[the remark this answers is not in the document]
aruniverse: ah, we re-export everything from the base pkg in the web. i guess this is ok. 
but the same should appy in desktop then, you should be able to import the providers from the desktop-viewer-react pkg then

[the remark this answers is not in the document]
achrysaetos: Oh... makes sense.

packages/apps/desktop-viewer-test/src/frontend/components/routes/ViewerRoute.tsx
```diff
@@ -35,10 +38,14 @@ export const ViewerRoute = ({ location }: ViewerRouteProps) => {
     <Viewer
       filePath={filePath}
       onIModelAppInit={onIModelAppInitialized}
-      defaultUiConfig={{
-        contentManipulationTools: { cornerItem: { hideDefault: true } },
-      }}
```
achrysaetos, line 40: Is contentManipulationTools deprecated? The type of defaultUIConfig is from ItwinViewerUi
image

achrysaetos: Should the desktop viewer look exactly like the web viewer? When I ran the desktop viewer, it didn't have any navigation tools either:
image

achrysaetos: Oh yeah, my bad, I haven't done git pull in a long time, that was why.

aruniverse: looks like you dropped your changes to the desktop side with the merge.

make sure you did:
```
git fetch --all
git merge origin/next
```
then resolve any conflicts.
look at the diff in your pr here to see what changes were made

aruniverse (approved):

PR #185, opened 2022-06-06
Initialization error fix

samejima-san:
fixing the initialization error on the web viewer when the user calls the imodelapp start up before the webviewer does.


packages/modules/desktop-viewer-react/src/services/Initializer.ts
```diff
@@ -107,8 +107,12 @@ export class DesktopInitializer {
           DesktopInitializer._initializing = false;
           DesktopInitializer._cancel = undefined;
         });
-    } else {
+    } else if (this._initializing) {
       this._initialized = Promise.resolve();
```
kckst8, line 111: lets change this to something like

`if (!this._initialized) {
       this._initialized = Promise.resolve();
} 
return this._initialized;
`
and use the same logic in both the web and desktop Initializers

packages/modules/desktop-viewer-react/src/services/Initializer.ts
```diff
@@ -107,8 +107,10 @@ export class DesktopInitializer {
           DesktopInitializer._initializing = false;
           DesktopInitializer._cancel = undefined;
         });
-    } else if (!this._initialized) {
-      this._initialized = Promise.resolve();
+    } else if (this._initializing) {
+      if(!this._initialized){
```
kckst8, line 111: ensure that you have the prettier extension installed and ensure that your files are formatted. Maybe [turn on format on save](https://linuxpip.org/vscode-format-on-save/).

samejima-san: it is installed. I am looking for where the issue could be.

kckst8 (commented):
your test is failing for me. You should run "npm test" from within the modules/web-viewer-react directory to ensure that tests pass (and "rush test" from any directory to ensure that tests in all packages pass before pushing commits)

aruniverse:
Closing in favor of https://github.com/iTwin/viewer/pull/210

PR #187, opened 2022-06-14
cleanup next

aruniverse:


johnnyd710 (changes requested):
LGTM! But I think you missed the `defaultUiConfig` mention in the desktop-viewer-react README ?

packages/modules/error-handling-react/tsconfig.json
```diff
@@ -1,11 +1,11 @@
 {
   "compilerOptions": {
-    "target":  "es2020",
+    "target": "es2017",
```
kckst8, line 3: are we still having issues supporting es2020?

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -62,7 +61,6 @@ const IModelLoader = React.memo(
     const { finalFrontstages, noConnectionRequired, customDefaultFrontstage } =
       useFrontstages(
         frontstages,
-        defaultUiConfig,
```
kckst8, line 65: do we have another way to hide the status bar? The CLI is currently using prop to hide it. The status bar provider just provides the items, but doesn't control visibility. Maybe we should just hide the bar when that provider isn't used?

packages/modules/error-handling-react/tsconfig.json
aruniverse: i just reverted this because we refactored this pkg out of the base viewer, so we dont use it anymore and i didnt want to make any new changes that we wouldnt have published. leaving the pkg as is as of our 2.x release

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
aruniverse: ah good catch, ill look into this

packages/modules/error-handling-react/tsconfig.json
kckst8: ah right. Should've paid closer attention to which package

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
aruniverse: the [StandardFrontstageProvider]() we use from core by default adds the navigation aid cube, the statusbar and the tool settings bar. you can optionally hide the navvigation aid and the status bar but not the tool settings. i can look into exposing an option to hide the tool settings bar in 3.3 but dont want to bump up the min core ver. I can re-add our own frontstage provider that forks what core has but adding that extra option. unofrutnately dont think there is an easy way to dynamically enable/disable these...
will continue investigating and update pr tomorrow

kckst8: There is a prop to hide the status bar  in the core front stage provider that I used in the cli pr. That part is fine. We just need a clean way to expose it if we remove this prop

aruniverse: youre ok with not being able to hide the ToolSettingsBar at all?

kckst8: For now I think so, but we should add that option in the core frontstage

aruniverse: added

kckst8 (approved):

PR #188, opened 2022-06-15
[next] fix tests, update monorepo tooling, add missing peers

aruniverse:
- fix tests
- update monorepo tooling (rush and pnpm vers to align with core repo)
- add missing peer deps to make sure theres always only ever 1 instance of core-frontend


packages/modules/desktop-viewer-react/package.json
```diff
@@ -23,7 +23,7 @@
     "rebuild": "npm run clean && npm run build"
   },
   "dependencies": {
-    "@itwin/viewer-react": "3.0.0-dev.9"
+    "@itwin/viewer-react": "workspace:*"
```
kckst8, line 26: curious, how does this work with rush publish? Is rush smart enough to know to replace it in the published package but not write back to the repo?

packages/modules/test-local-extension/package.json
```diff
@@ -13,14 +13,21 @@
...
+    "@itwin/core-geometry": "^3.2.0",
+    "@itwin/core-orbitgt": "^3.2.0",
+    "@itwin/core-quantity": "^3.2.0",
+    "@itwin/webgl-compatibility": "^3.2.0",
     "esbuild": "^0.13.13",
+    "npm-run-all": "^4.1.5",
     "rimraf": "^3.0.2",
     "serve": "^13.0.2",
-    "npm-run-all": "^4.1.5"
+    "typescript": "~4.4.0"
```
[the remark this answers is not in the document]
aruniverse: its actually using the same ver as before, i just alphabetized the deps in addition to adding the missing peers.

using ~ instead of ^ to make sure the entire monorep (our and cores) all use the same ver of ts
we are a few versions behind so we can update ts across this monorepo before next or after, up to you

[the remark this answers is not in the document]
kckst8: ok yeah I guess to keep with core I guess. Really we should never be using ~. Didnt even realize we were doing that

packages/modules/desktop-viewer-react/package.json
kckst8: Ok trust you there :-)

kckst8 (approved):

packages/modules/test-local-extension/package.json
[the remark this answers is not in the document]
aruniverse: ts is weird, where every minor release its their beta release, and their ga/official release is of the .2 patch for that minor

i guess it wouldnt hurt to use ^ as most things id worry about in a ts upgrade would probably need to be opted in in the tsconfig file, but not 100% sure. those guys do have breaking changes in minor releases too

packages/modules/desktop-viewer-react/package.json
aruniverse: ill test it out locally with a dry run before merging into master

aruniverse: tested with a next release and works as expected

PR #191, opened 2022-06-16
fix desktop-viewer@next template for node16

aruniverse:
in node 16, the newer versions of npm (7+) are stricter on peer deps, so the desktop templates failed with: 
image
updated from @reach/router@1 -> react-router@6, see:
- https://reactrouter.com/docs/en/v6/upgrading/reach
- https://reactrouter.com/docs/en/v6/upgrading/v5

also made reality-data-client a real dep in the base viewer pkg


johnnyd710 (approved):
LGTM!

aruniverse:
merging with admin priveleges. john tested locally on windows and runs fine on his end

PR #203, opened 2022-07-15
use globalThis instead of process.env to set env at runtime

aruniverse:
configure env at runtime instead of only at buildtime

see related pr: https://github.com/imodeljs/create-react-app/pull/70

### DIFF
--- a/packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
+++ b/packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
@@ -6,7 +6,7 @@
 import { ColorTheme } from "@itwin/appui-react";
 import { BrowserAuthorizationClient } from "@itwin/browser-authorization";
 import {
-  LocalExtensionProvider,
+  // LocalExtensionProvider,
   RemoteExtensionProvider,
 } from "@itwin/core-frontend";
 import {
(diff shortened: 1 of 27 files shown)


johnnyd710 (approved):
nice so we don't have to rebuild if we forget to set the dev environment? :)

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
```diff
@@ -142,10 +142,10 @@ const ViewerHome: React.FC = () => {
           new MeasureToolsUiItemsProvider(),
         ]}
         extensions={[
-          new LocalExtensionProvider({
-            manifestPromise: LocalExtension.manifestPromise,
-            main: LocalExtension.main,
-          }),
+          // new LocalExtensionProvider({
```
kckst8, line 145: why is this commented?

packages/apps/web-viewer-test/tsconfig.json
```diff
@@ -13,8 +17,10 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx",
+    "jsx": "react",
```
kckst8, line 20: isn't react-jsx the new value (React 17)?

aruniverse: this is one of those changes made by cra, ill revert

aruniverse: image

actually looks like this is the right one we want for react 17

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
aruniverse: had run into this earlier, had to clean and rebuild a few times before it worked. 
reverted.

image

kckst8 (approved):

PR #207, opened 2022-07-22
update codeowners

kckst8:
left @aruniverse and @calebmshafer but removed myself

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,8 +1,13 @@
 # Each line is a file pattern followed by one or more owners.
+
 # The owners will be requested for
+
 # review when someone opens a pull request.
+
 # See https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
 
 # These owners will be the default owners for everything in
+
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,8 +1,13 @@
...
 # review when someone opens a pull request.
+
 # See https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
 
 # These owners will be the default owners for everything in
+
 # the repo, unless a later match takes precedence.
-*       @kckst8 @aruniverse @calebmshafer
+
+-       @iTwin/itwinjs-core-ui @aruniverse @calebmshafer
```
aruniverse, line 13: does it make more sense to just add the viewer team, and update who is apart of it?

aruniverse (approved):

.github/CODEOWNERS
kckst8: @NancyMcCallB asked for this team (I guess since the members would all be the same). @NancyMcCallB, would you rather maintain a second team specifically for the Viewer?

Issue #157, opened 2022-03-29
better handle initialization errors


kckst8:
The useWebViewerInitializer hook assumes that the "initialized" promise is defined in the WebInitializer class. The issue is if a user has already initialized IModelApp, it will not be defined, due to no "else" condition here:

` if (!IModelApp.initialized && !this._initializing) {`

this results in a "cannot find "then" of undefined" error instead of returning the promise if this._initializing and rejecting the promise with a meaningful error (let the user know to use the useWebViewerInitializer hook instead of calling IModelApp.startup themselves) if IModelApp.Initialized.

the same could occur in the desktop package's DesktopInitializer class, so it should be fixed in both.

kckst8:
@samejima-san ping me before starting work on this

PR #210, opened 2022-08-02
Better handle initialization errors

achrysaetos:
Closes https://github.com/iTwin/viewer/issues/157

### DIFF
--- a/common/changes/@itwin/desktop-viewer-react/leck-initialize_2022-08-02-22-04.json
+++ b/common/changes/@itwin/desktop-viewer-react/leck-initialize_2022-08-02-22-04.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/desktop-viewer-react",
-      "comment": "Fixes WebInitializer.initialized error if IModelApp is already initialized",
+      "comment": "Fixes DesktopInitializer.initialized error if IModelApp is already initialized",
       "type": "patch"
     }
   ],
--- a/common/changes/@itwin/desktop-viewer-react/leck-initialize_2022-08-02-22-04.json
+++ b/common/changes/@itwin/desktop-viewer-react/leck-initialize_2022-08-02-22-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/desktop-viewer-react",
+      "comment": "Fixes WebInitializer.initialized error if IModelApp is already initialized",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/desktop-viewer-react"
+}
--- a/packages/modules/desktop-viewer-react/src/services/Initializer.ts
+++ b/packages/modules/desktop-viewer-react/src/services/Initializer.ts
@@ -112,6 +112,10 @@ export class DesktopInitializer {
         this._initialized = Promise.resolve();
       }
       return this._initialized;
+    } else {
+      throw new Error(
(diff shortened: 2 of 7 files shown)


common/changes/@itwin/desktop-viewer-react/leck-initialize_2022-08-02-22-04.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/desktop-viewer-react",
+      "comment": "Fixes WebInitializer.initialized error if IModelApp is already initialized",
```
aruniverse, line 5: DesktopInitializer

packages/modules/desktop-viewer-react/src/services/Initializer.ts
```diff
@@ -107,8 +107,11 @@ export class DesktopInitializer {
           DesktopInitializer._initializing = false;
           DesktopInitializer._cancel = undefined;
         });
-    } else {
-      this._initialized = Promise.resolve();
+    } else if (this._initializing) {
```
BibekSingh09, line 110: What would happen if _initializing is false, we don't have any condition to handle the case - 

image

achrysaetos: good catch.

BibekSingh09 (approved):
I have tested, this fix seems to be working for me.

kckst8 (changes requested):
The thing that is missing here is another else if before throwing the error

`else if (this._initializing) {
    return this._initialized;
}`

Basically if another initialization already started we want to return the promise if that makes sense. Another way might be to check IModelApp.initialized and just throw at the beginning and then you just have 2 conditions in the conditional.

Basically there are 3 cases that need to be handled:

1. Nothing is initialized, proceed
2. IModelApp is not yet initialized but the process has been started, don't start a second initialization but also dont throw. Just let the first complete.
3. IModelApp already initialized. Throw an error directing the user to the hook instead.

achrysaetos:
@kckst8 Makes sense! tysm

kckst8 (approved):

BeAliAslam:
There is a problem in version 3.0.7 due to recent changes (via this PR) in src/services/initializer.ts>> startWebViewer function. 
The if (IModelApp.initialize) would fail if we re-render viewer from host by changing some viewerprop as useWebViewerInitializer.tsx useeffect gets kicked in and calls WebInitializer.startWebViewer. 

One use case i have is that if you add say another 'additional rpc interface' via additionalRpcInterfaces prop of viewer in host by say adding the interface in onIModelAppInit callback, the viewer throws error in startWebViewer function. 

In sample app, this can be reproduced below

App.tsx
   const [additionalRpcInterfaces, setAdditionalRpcInterfaces] = useState<RpcInterfaceDefinition<RpcInterface>[]>([]);
...
const onIModelConnected = ()=>{
...
const additionalRpcsLocal = [...additionalRpcInterfaces];

  additionalRpcsLocal.push(DevToolsRpcInterface);

setAdditionalRpcInterfaces([...additionalRpcsLocal])
}

<Viewer
        iTwinId={iTwinId ?? ""}
      ...
        additionalRpcInterfaces={additionalRpcInterfaces}
        onIModelAppInit={onIModelAppInit}
...

@achrysaetos @DarrinDoherty

aruniverse:
@BeAliAslam , ive opened a pr to revert this change for the time being, https://github.com/iTwin/viewer/pull/217

Issue #211, opened 2022-08-10
Backstage button missing


aruniverse:
user is using UiProvider to provideBackstageItems but the backstage button is missing in viewer-react@3.x

aruniverse:
fixed by https://github.com/iTwin/viewer/pull/291

PR #212, opened 2022-08-12
Expose cornerButton option

achrysaetos:
Closes https://github.com/iTwin/viewer/issues/211.

### DIFF
--- a/packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
+++ b/packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { ColorTheme } from "@itwin/appui-react";
+import { BackstageAppButton, ColorTheme } from "@itwin/appui-react";
 import { BrowserAuthorizationClient } from "@itwin/browser-authorization";
 import {
   LocalExtensionProvider,
@@ -24,6 +24,7 @@ import {
 } from "@itwin/tree-widget-react";
 import type { ViewerBackstageItem } from "@itwin/web-viewer-react";
 import {
+  BackstageItemsProvider,
   Viewer,
   ViewerContentToolsProvider,
   ViewerNavigationToolsProvider,
@@ -111,6 +112,8 @@ const ViewerHome: React.FC = () => {
     },
   ];
 
+  const cornerButton = <BackstageAppButton />;
+
   return (
     <div style={{ height: "100vh" }}>
       <Viewer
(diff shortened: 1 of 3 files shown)


aruniverse (approved):
wheres the changes in the test app?

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
```diff
@@ -152,6 +156,7 @@ const ViewerHome: React.FC = () => {
           }),
         ]}
         backstageItems={backstageItems}
```
aruniverse, line 158: do we need this and provider? i guess prefer the provider over passing directly into the viewer

achrysaetos: Yes, it needs both: if the backstageItems property is removed, the button can't be clicked. If the provider is removed, the button can be clicked but the panel will be empty.

aruniverse: ok this seems wrong imo... lets hold off on completing this pr then

aruniverse: feels wrong to have both, maybe if you pass in backstageItems, it auto adds the BackstageProvider under the hood?

raplemie: In the latest code, the BackstageItemsProvider is no longer used... and from what I can see backstageItems currently only serves to start the BackstageComposer, but it does not provide it the backstage items... I guess we screwed up somewhere in the refactor :)

Below is the only place where we "do" something with backstageItems value, other than passing it on down the chain.
image

aruniverse: I was supposed to remove this prop altogether in the 3.0 release, forgot to scrub TODO comments :/
https://github.com/iTwin/viewer/blob/master/packages/modules/viewer-react/src/components/iModel/IModelViewer.tsx#L20

aruniverse: I know we've had previous discussions with the team about not pushing backstages because its really never caught on and has been very confusing. Since the user can provide their own custom frontstage wonder if we should do anything about this. Maybe we should chat about this in next week ui/viewer call?

raplemie: If anything, we should validate what the <BackstageComposer /> does, and if it really need to be removed if there is not backstage items provided. It feels like it should simply always be there and handle UIProvided items, and it would be up to the user to provide its own frontstage that includes the backstage button (where they have access to the full props anyway)

packages/modules/viewer-react/src/types.ts
```diff
@@ -207,5 +207,5 @@ export interface ViewerDefaultStatusbarItems {
 
 export type ViewerDefaultFrontstageConfig = Pick<
   StandardFrontstageProps,
-  "hideNavigationAid" | "hideStatusBar" | "hideToolSettings"
+  "hideNavigationAid" | "hideStatusBar" | "hideToolSettings" | "cornerButton"
```
aruniverse, line 210: since youre expanding this api, add an example of this usage in the test app

aruniverse:
replaced by https://github.com/iTwin/viewer/pull/291

PR #213, opened 2022-08-15
Create a widget to query ECSQL strings

achrysaetos:


common/changes/@itwin/viewer-react/leck-ecsql-widget_2022-08-15-20-49.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/viewer-react",
```
aruniverse, line 4: we dont want this in the viewer modules itself, just in the test app

achrysaetos: Do you mean move the provider to inside the test app instead of in viewer-react?

aruniverse: @achrysaetos Idk where youre comment went, i dont see it in the pr
image

But yes. This should not be apart of the viewer-react pkg. Just needs to be apart of the test app. So can just move the code from the module into the app

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { UiFramework } from "@itwin/appui-react";
```
aruniverse, line 6: Would prefer you not use `UiFramework` and instead use `useActiveIModelConnection` hook instead to obtain the iModelConnection

aruniverse:
> 1. the table becomes very slow if the number of results exceeds 5000 rows

You probably want to be using a [Virtualized Table](https://itwin.github.io/iTwinUI-react/?path=/story/core-table--virtualized)

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+import {
+  Button,
+  ExpandableBlock,
+  LabeledTextarea,
+  Table,
+} from "@itwin/itwinui-react";
+import React, { useState } from "react";
+
+export default function EcsqlWidget() {
+  const [input, setInput] = useState<string>("");
```
aruniverse, line 17: you shouldnt need to type the state explicitly for strings and booleans, those should be able to be inferred by the literals you initialize the state with

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const [rows, setRows] = useState<Record<string, string>[]>([]);
+  const [headers, setHeaders] = useState<string[]>([]);
+  const [isloading, setLoading] = useState<boolean>(false);
```
aruniverse, line 20: prefer `isLoading` and `setIsLoading`

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+    try {
+      setRows([]);
+      setHeaders([]);
+      setLoading(true);
+      const results = UiFramework.getIModelConnection()!.query(
+        input,
+        undefined,
+        { rowFormat: QueryRowFormat.UseJsPropertyNames }
+      );
+      const rows = [];
```
aruniverse, line 33: be careful with var names; make sure to look at the scope.
youre reusing the same names in the scope of this function but also at the component level

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+          .forEach((key) => {
+            row[key] = obj[key].toString();
+          });
+        // Case 2: Need to stringify objects to display them in the table
+        Object.keys(row)
+          .filter((key) => typeof row[key] === "object")
+          .forEach((key) => {
+            row[key] = JSON.stringify(obj[key]);
+          });
+        // Case 3: If the cell content is too long, put it inside an ExpandableBlock
```
aruniverse, line 50: what does this actually look like?

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const columns = React.useMemo(
+    () => [
+      {
+        Header: "Header name",
+        columns: headers.map((str) => ({
+          id: str,
+          Header: str,
+          accessor: str,
+          width: 300,
```
aruniverse, line 88: whats the default?

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+        })),
+      },
+    ],
+    [headers]
+  );
+  const data = rows;
```
aruniverse, line 94: is this necessary? need an entire diff var to pass to the table?

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  return (
+    <div style={{ height: "100%", display: "flex", flexFlow: "column" }}>
```
aruniverse, line 97: for now inline styles are fine, and you can leave this as is, but just a heads up we prefer using scss files

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
```diff
@@ -0,0 +1,141 @@
...
+        <LabeledTextarea
+          label=""
+          message={error ? error : data.length.toString() + " results"}
+          value={input}
+          onChange={(e) => setInput(e.target.value)}
+          status={error ? "negative" : undefined}
+          rows={1}
+          style={{ flexGrow: 1 }}
+        />
+        <Button
```
aruniverse, line 116: wonder if we need a button or instead have a button inside the label, take a look at the different components and options https://itwin.github.io/iTwinUI-react/?path=/story/input-labeledinput--hybrid-layout-with-button&args=label:

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
aruniverse: what is the type for row here when this occurs? i dont think it matches what you had originally typed it as

packages/modules/viewer-react/src/components/app-ui/providers/EcsqlWidgetProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { StagePanelLocation, StagePanelSection } from "@itwin/appui-abstract";
+import React from "react";
+
+import EcsqlWidget from "../EcsqlWidget";
+
+export class EcsqlWidgetProvider implements UiItemsProvider {
+  public readonly id = "EcsqlWidgetProvider";
+
+  public provideWidgets(
+    stageId: string,
```
aruniverse, line 19: Take a look at what we do for stageId and stageUsage in the other providers in viewer-components-react

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
achrysaetos: Looks like this:
image

When expanded:
image

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
achrysaetos: The default is where each column takes up the same proportion of the screen, so there's no horizontal scrollbar:
image

With width=300:
image

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
achrysaetos: I'm currently using a [labeled textarea](https://itwin.github.io/iTwinUI-react/?path=/story/input-labeledtextarea--negative) so that the user can resize the textbox if the query gets too long. A labeledInput wouldn't need the button, but also wouldn't have the resize functionality.

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
aruniverse: you missed `isLoading`

packages/modules/viewer-react/src/components/app-ui/EcsqlWidget.tsx
aruniverse: still feels a little weird for me, have you looked at what the standard pattern we use is?
do we just show as much text as we can before showing an ellipses for overflow, and then on hover show the full text in a tool tip or something?

achrysaetos: iModelConsole shows as much as it can in a cell before showing an ellipses for overflow, and then clicking the cell expands the whole row to display the full text. The Table from core doesn't have anything like that, I think, there's just a button to expand the whole row
image

achrysaetos: Or maybe I just have to customize it a bit more.

achrysaetos:
> > 1. the table becomes very slow if the number of results exceeds 5000 rows
> 
> You probably want to be using a [Virtualized Table](https://itwin.github.io/iTwinUI-react/?path=/story/core-table--virtualized)

Unfortunately, the virtualized table loses the horizontal scrollbar if there are too many columns, and it can't be added back without some (hacky?) css ☹️. Also for some long strings, the virtualized table will only write the visible portion to the DOM, which results in it trying to constantly format each string to fit inside its cell, leading to a jittery experience imo.

The performance gain is very important though...

packages/apps/web-viewer-test/src/providers/EcsqlWidget.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+export default function EcsqlWidget() {
+  const [input, setInput] = useState("");
+  const [data, setData] = useState<Record<string, string | React.ReactNode>[]>(
+    []
+  );
+  const [headers, setHeaders] = useState<string[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState("");
+  const iModelConnection = useActiveIModelConnection() as IModelConnection;
```
aruniverse, line 25: shouldnt need to do a cast here, its of type `IModelConnection | undefined` which is ok

packages/apps/web-viewer-test/src/providers/EcsqlWidget.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const getData = async () => {
+    try {
+      setData([]);
+      setHeaders([]);
+      setIsLoading(true);
+      const results = iModelConnection.query(input, undefined, {
```
aruniverse, line 32: do a null check for iModelConnection right before the try block to make sure its defined,

packages/apps/web-viewer-test/src/providers/EcsqlWidget.tsx
achrysaetos: If I don't do a cast, I might need the `!` operator, right?
image

packages/apps/web-viewer-test/src/providers/EcsqlWidget.tsx
```diff
@@ -0,0 +1,147 @@
...
+            row[key] = (
+              <ExpandableBlock title="..." size="small" styleType="borderless">
+                {row[key]}
+              </ExpandableBlock>
+            );
+          });
+
+        rows.push(row);
+        for (const key of Object.keys(row)) {
+          if (!columnHeaders.includes(key)) {
```
aruniverse, line 67: well wouldnt they always all not be in columnheaders to begin with?
re-read this block of code here

packages/apps/web-viewer-test/src/providers/EcsqlWidget.tsx
```diff
@@ -0,0 +1,147 @@
...
+        }
+      }
+      setData(rows);
+      setHeaders(columnHeaders);
+      setIsLoading(false);
+      setError("");
+    } catch (err: any) {
+      setIsLoading(false);
+      setError(err.message);
+    }
```
aruniverse, line 79: nit, no need to change, but you could add a `finally` block where you do `setIsLoading(false)`

aruniverse (commented):
Good work! We'll take it over once you go back to school. Unless you want to continue contributing as an OpenSource Collaborator ;)

packages/apps/web-viewer-test/src/providers/EcsqlWidget.tsx
aruniverse: look at my next comment, do a null check earlier

aruniverse: what happens if it was actually undefined and you cast it to an iModelConnection?

achrysaetos: You're right, that would be bad.

packages/apps/web-viewer-test/src/providers/EcsqlWidget.tsx
achrysaetos: Hmm, I'm pretty sure this is right. I wanted columnHeaders to be the superset of all keys in each row, because sometimes each row might have different keys. So every time columnHeaders encounters a new unique key, it should append the new key. columnHeaders is instantiated before we start iterating through each row.

PR #216, opened 2022-09-06
Update props needed for Blank Connection Viewer

ben-polinsky:


packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -91,8 +96,8 @@ const IModelLoader = React.memo(
       );
       let imodelConnection: IModelConnection | undefined;
       // create a new imodelConnection for the passed project and imodel ids or local file
-      if (blankConnection) {
-        imodelConnection = BlankConnection.create(blankConnection);
+      if (iTwinId && !iModelId) {
```
calebmshafer, line 99: should this include the blankConnection as well? Or should we try and remove the need for the blankConnection as well?

ben-polinsky: Considered it, but decided to play it safe and allow an override for now.

ben-polinsky: Could extract the overrides to separate/different props...

aruniverse: actually doesnt really make sense for check the blankConnection because its optional now

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -197,4 +202,23 @@ const IModelLoader = React.memo(
   }
 );
 
+/**
+ * Create a blank connection with default props
+ * @param blankConnectionProps?
+ * @returns BlankConnection
+ */
+const createBlankConnection = (
```
aruniverse, line 210: if we're going to go with the approach where you only need an iTwinId then we should pass it as another arg here,
https://www.itwinjs.org/reference/core-frontend/imodelconnection/blankconnectionprops/

ben-polinsky: Yes, makes sense. Good catch.

aruniverse: idk if the arg should be required, because what if the user provides blankConnection instead?
i guess it depends on what we do with that prop, and if we can remove it w/o a breaking change

ben-polinsky: Was working under the assumption that iTwinId was always required, but I see now.

ben-polinsky: I can just make the blankConnection required again, if that is easiest

packages/modules/viewer-react/src/tests/services/iModelViewerHelper.test.ts
```diff
@@ -0,0 +1,226 @@
...
+import * as iModelService from "../../services/iModel/IModelService";
+import {
+  createBlankConnection,
+  getConnectionType,
+  openConnection,
+} from "../../services/iModel/iModelViewerHelper";
+import type { RequiredViewerConnectionProps } from "../../types";
+import { ConnectionType } from "../../types";
+
+jest.mock("@itwin/core-frontend", () => {
```
ben-polinsky, line 20: I really wanted to extract this and use it in the BaseInitializer test, but had issues with jest hoisting/initializing modules.

packages/modules/viewer-react/src/types.ts
```diff
@@ -134,20 +141,59 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
 
 export interface ConnectedViewerProps {
   iTwinId: string;
-  iModelId: string;
+  iModelId?: string;
```
aruniverse, line 144: i think we should leave this as required, as we need both itwin and imodel ids to open a remote connection model

packages/modules/viewer-react/src/types.ts
```diff
@@ -134,20 +141,59 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
 }
-
 export interface FileViewerProps {
   /** Path to local snapshot or briefcase */
   filePath: string;
 }
 
-export interface BlankViewerProps {
-  blankConnection: BlankConnectionProps;
+export type BlankViewerProps = {
```
aruniverse, line 152: wondering if we should split this type further down and have it be a union of the deprecated way and the updated/recommended usage. youre kina already doing this 

as is rn, BlankViewerProps === BlankViewerPropsWithoutDeprecation (ignoring the deprecated jsdoc tag)

packages/modules/viewer-react/src/types.ts
```diff
@@ -134,20 +141,59 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+  extents?: Range3dProps;
   blankConnectionViewState?: BlankConnectionViewState;
+};
+
+export interface BlankConnectionInitializationProps {
+  iTwinId?: string;
+  blankConnectionProps: BlankConnectionProps;
+}
+
+export interface RequiredViewerConnectionProps {
```
aruniverse, line 172: is it really _required_ if everything is optionally defined?

packages/modules/viewer-react/src/types.ts
```diff
@@ -134,20 +141,59 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+  iTwinId?: string;
+  iModelId?: string;
+  filePath?: string;
+  blankConnection?: BlankConnectionProps;
+  extents?: Range3dProps;
+  location?: Cartographic | EcefLocationProps;
 }
 
+export enum ConnectionType {
+  None,
```
aruniverse, line 182: this really shouldnt be a valid case, and i think the way the types were setup if you didnt satisfy the XOR union it would yell at you that youre missing props

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
```diff
@@ -0,0 +1,91 @@
...
+export const createBlankConnection = ({
+  iTwinId,
+  blankConnectionProps,
+}: BlankConnectionInitializationProps) =>
+  BlankConnection.create({
+    iTwinId,
+    ...blankConnectionProps,
+  });
+
+export const getConnectionType = ({
```
aruniverse, line 33: i was on a TS craze when i did my refactor for 3.0; wondering if we should use type guards instead

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
```diff
@@ -0,0 +1,91 @@
...
+    (blankConnection && iTwinId) ||
+    (extents && location && iTwinId)
+  ) {
+    return ConnectionType.Blank;
+  }
+
+  return ConnectionType.None;
+};
+
+export const openConnection = async (
```
aruniverse, line 58: we should avoid using non null assertions imo, i know everything is defined here because of connectionType, but i think we can get TS to correctly infer the types if we use type guards

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
ben-polinsky: Yep, I can clean this up.

packages/modules/viewer-react/src/types.ts
ben-polinsky: Removed ConnectionType. Was adding it to aid in discriminated unions, but using your `XOR` was fine.

packages/modules/viewer-react/src/types.ts
ben-polinsky: Revamped.

packages/modules/viewer-react/src/types.ts
ben-polinsky: Also reworked.

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
ben-polinsky: Reworked.

packages/modules/viewer-react/src/types.ts
ben-polinsky: Done.

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,21 +141,53 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+
+export type BlankConnectionInitializationProps = {
+  iTwinId?: string;
+  blankConnectionProps: BlankConnectionProps;
+};
 
-export interface BlankViewerProps {
-  blankConnection: BlankConnectionProps;
-  blankConnectionViewState?: BlankConnectionViewState;
+export enum ConnectionType {
```
aruniverse, line 187: i think we can delete this, especially as we are able to infer the type

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,21 +141,53 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
-  iTwinId: string;
+export type RequiredViewerProps = XOR<
+  XOR<ConnectedViewerProps, FileViewerProps>,
+  BlankViewerProps
+>;
+export type ViewerProps = RequiredViewerProps & ViewerCommonProps;
+export type ViewerLoaderProps = RequiredViewerProps & LoaderProps;
+export type ConnectedViewerProps = {
+  iTwinId: GuidString;
   iModelId: string;
```
aruniverse, line 152: i know GuidString is just an alias for string, but i guess might as well change iModelId as well if we're changing iTwinId

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,21 +141,53 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+export type FileViewerProps = {
   /** Path to local snapshot or briefcase */
   filePath: string;
-}
+};
+
+export type BlankViewerProps = XOR<
+  XOR<
+    {
+      /** @deprecated specify location and extents instead. */
```
aruniverse, line 163: thoughts on breaking these into their own types, so we can mark the whole subtype as deprecated, and not just the blankConnection property?

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,21 +141,53 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+      blankConnection: PropRequired<BlankConnectionProps, "iTwinId">;
+      blankConnectionViewState?: BlankConnectionViewState;
+    },
+    {
+      /** @deprecated specify location and extents instead. */
```
aruniverse, line 168: wonder if we need this one at all; it might be an unecessary step in between the deprecated api and where we're pushing the api fwd to; might as well have consumers just update to the preferred api.

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,21 +141,53 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
   /** Path to local snapshot or briefcase */
   filePath: string;
-}
+};
+
+export type BlankViewerProps = XOR<
+  XOR<
+    {
+      /** @deprecated specify location and extents instead. */
+      blankConnection: PropRequired<BlankConnectionProps, "iTwinId">;
```
aruniverse, line 164: i like this, but does this make it a breaking change?

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,21 +141,53 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+      blankConnectionViewState?: BlankConnectionViewState;
+    },
+    {
+      /** @deprecated specify location and extents instead. */
+      blankConnection: BlankConnectionProps;
+      blankConnectionViewState?: BlankConnectionViewState;
+      iTwinId: GuidString;
```
aruniverse, line 171: if we keep this, should this be optional?

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,21 +141,53 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
   /** array of iTwin.js Extensions */
   extensions?: ExtensionProvider[];
 }
-
-export interface ConnectedViewerProps {
-  iTwinId: string;
+export type RequiredViewerProps = XOR<
```
aruniverse, line 144: maybe we should relook at whats being exported; not a big deal; but everythign here is re-esported at the top level pkgs, and wonder if we should be more aware of what our public api/types are

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (options.iTwinId && options.iModelId) {
+    return await openRemoteIModel(
+      options.iTwinId,
+      options.iModelId,
+      options.changeSetId
+    );
+  }
+
+  if (options.filePath) {
+    console.log("openming local i model");
```
aruniverse, line 47: opening

packages/modules/viewer-react/src/components/BaseViewer.tsx
```diff
@@ -41,7 +33,7 @@ export const BaseViewer = ({
   const accessToken = useAccessToken();
   return (
     <ErrorBoundary>
-      {("filePath" in loaderProps && loaderProps.filePath) || accessToken ? (
+      {loaderProps.filePath || accessToken ? (
```
aruniverse, line 36: surprised no ts errors here, nice

aruniverse (commented):
looking good, i need to  test it out locally. the clever use of typing is helpful (imo), but sometimes the warnings in the IDE and error-overlay can get ugly, so just want to check that.

we will need to update our docs (the readmes) before this pr goes through with updated blank viewer prop changes

packages/modules/viewer-react/src/types.ts
ben-polinsky: Yeah... I think I'll go back to string.

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
```diff
@@ -69,25 +69,27 @@ export const openConnection = async (
   return;
 };
 
-export const missingViewerConnectionProps = (
-  props: Partial<RequiredViewerProps>
-): boolean => {
+export const gatherRequiredViewerProps = (
```
ben-polinsky, line 72: @aruniverse Now relying on your `ModelLoaderProps` as Viewer input, but narrowing down to required props here. My last attempt at creating some tighter types. If you're not a fan, we can revisit when `blankConnectionProps` are removed.

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,22 +138,43 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
   /** array of iTwin.js Extensions */
   extensions?: ExtensionProvider[];
 }
+export type RequiredViewerProps = XOR<
+  XOR<ConnectedViewerProps, FileViewerProps>,
+  BlankViewerProps
+>;
+
+export type ModelLoaderProps = Partial<
```
aruniverse, line 146: I think instead of doing 
```ts
Partial<ConnectedViewerProps & FileViewerProps & BlankViewerProps> & ...
```
using a union like this is more powerful:
```ts
(ConnectedViewerProps | FileViewerProps | BlankViewerProps) & ...
```

TS is able to better infer the union type, and the intellisense you get is much stronger imo
[here's an exmaple](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gOwRAxsCATAalAvFAbwFgAoKKAS2AHcKEAuKAZ2ACc6BzAblPIoFs4GCABtGLdgg6kAvqVKhIUAGIUREHPmJkoAMzUQACgENgAC3FtOs+SUXQAQiOMIA1psK8oAI2dvGBHIkQQrg0PD8ggh4njrGjAgArvzeEKwANF7ejN5wcOouNiShSgDyIhgxABTwSKjo2FAAPioGOC1OLu4AlFAAZLBwkYi29lAAchDUMSaswBTGIgA8tchomDgDquqbUJ1uWAB8-YPDCLYoiCxQcIzllVpe8VAAjJk62VBsiRDvfDR0RgAIiB7yKpEuCGuDAmUxi2nIzzeWUY31+XiotBhILBwSAA)

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,22 +138,43 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+  ConnectedViewerProps & FileViewerProps & BlankViewerProps
+> &
+  LoaderProps;
+
+export type ViewerProps = RequiredViewerProps & ViewerCommonProps;
```
aruniverse, line 151: not the exact type from before, but its narrower so its a :thumbsup:

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
```diff
@@ -0,0 +1,102 @@
+/*---------------------------------------------------------------------------------------------
```
aruniverse, line 1: nit: for consistency, could you export this file in the barrel in this dir? and then consume exports from here from that barrel in the other files?

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -19,182 +19,152 @@ import {
...
+      viewCreatorOptions,
+      blankConnectionViewState,
+    });
+  useUiProviders(uiProviders);
+  useTheme(theme);
+
+  const getModelConnection = useCallback(async (): Promise<
+    IModelConnection | undefined
+  > => {
+    const sanitizedProps = gatherRequiredViewerProps(viewerProps);
```
aruniverse, line 62: nit: rename to requiredProps or requiredConnectionProps or something?

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
aruniverse: i like this a lot more

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  if (
+    blankConnection?.iTwinId ||
+    (blankConnection && iTwinId) ||
+    (extents && location && iTwinId)
+  ) {
+    return { blankConnection, iTwinId, extents, location };
+  }
+
+  return false;
```
aruniverse, line 101: should we return undefined instead? does that make more sense than false?

common/scripts/.eslintrc.ts.json
```diff
@@ -8,6 +8,6 @@
     "@typescript-eslint/prefer-optional-chain": "warn",
     "@typescript-eslint/prefer-nullish-coalescing": "warn",
     "react-hooks/exhaustive-deps": "warn",
-    "deprecation/deprecation": "error"
+    "deprecation/deprecation": "warn"
```
aruniverse, line 11: lets revert this

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
```diff
@@ -0,0 +1,102 @@
...
+      iTwinId: options.iTwinId,
+      blankConnectionProps: {
+        extents: options.extents,
+        location: options.location,
+        name: "Blank Connection",
+      },
+    });
+  }
+
+  if (options.blankConnection) {
```
aruniverse, line 62: once we revert the eslint change, we'll need to explicitly disable deprecation for this line, and the others, or maybe just the entire file

packages/modules/viewer-react/src/types.ts
```diff
@@ -131,22 +138,43 @@ export interface ViewerInitializerParams extends ViewerIModelAppOptions {
...
+  /** @deprecated specify location and extents instead. */
+  blankConnection?: BlankConnectionProps;
   blankConnectionViewState?: BlankConnectionViewState;
-}
+  location?: Cartographic | EcefLocationProps;
+  extents?: Range3dProps;
+  iTwinId?: string;
+};
+
+export type BlankConnectionInitializationProps = {
```
aruniverse, line 174: should we just move this into iModelViewHelper since no one else uses this?

common/changes/@itwin/viewer-react/benp-optional-imodel-id_2022-09-06-18-35.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/viewer-react",
+      "comment": "Remove need for user-supplied iModelId",
+      "type": "patch"
```
aruniverse, line 6: maybe make this a minor bump?
and change the comment so its says something like blank connection will be created if iTwinId is supplied and iModelId is not? might need to work on the wording there

aruniverse: might want to update the other changelog entry as well

aruniverse (commented):
looking good, didnt review test code (i trust you there), havent ran locally yet

packages/modules/viewer-react/src/types.ts
ben-polinsky: Yes, thanks makes sense.

ben-polinsky: Oh the issue is that we need the possibility of accepting any of the props, and the union restricts the props to one of the types:

```ts
 export type ModelLoaderProps = (
  ConnectedViewerProps | FileViewerProps | BlankViewerProps
 ) & LoaderProps;
```
image

aruniverse: so we cant destructure completely but would have to do props.iTwinId && props.iModelId or props.filePath instead?

Idk maybe not worth it, ill leave it up to you to resolve. thx for calling this out

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
ben-polinsky: Yep - will do. Leftover from when this was `hasRequiredConnectionProps`

packages/modules/viewer-react/src/types.ts
ben-polinsky: Even worse, will have to do `if ("filePath" in props)...`

ben-polinsky: I think we can circle back with some better types once we make some breaking changes

aruniverse: yea that sounds ok to me

common/scripts/.eslintrc.ts.json
ben-polinsky: Figured this out.. for some reason cannot ignore the deprecation errors if I leave the deprecation/deprecation rule here. If I move it out to the base eslint config, then it works 🤷‍♂️

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
ben-polinsky: Done.

packages/modules/viewer-react/src/types.ts
ben-polinsky: Done

packages/modules/viewer-react/src/services/iModel/iModelViewerHelper.ts
ben-polinsky: done

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
ben-polinsky: done

common/changes/@itwin/viewer-react/benp-optional-imodel-id_2022-09-06-18-35.json
ben-polinsky: Holding this open until I test changes in the test-app. Will add change then.

ben-polinsky: Alright tests out fine. If you have time, take another pass at my responses/corrections and LGTM (let's get this merged :))

CONTRIBUTING.MD
```diff
@@ -60,8 +60,8 @@ Every change must be described with a change log: Run `rush change` on your comm
 1. Make source code changes on a new Git branch.
 2. Add unit tests.
 3. Ensure unit tests pass when run locally: `rush test`.
-4. Test changes with the appropriate development app for your changes (web-viewer-test, desktop-viewer-test).
-   - From root directory - `rushx build` (alternatively, run `rushx start`in each project folder that you are editing as well as the development app folder to watch for changes and re-compile).
+4. Test changes with the appropriate development app for your changes (web-viewer-test, cdn-viewer-test, desktop-viewer-test).
```
aruniverse, line 63:
```suggestion
4. Test changes with the appropriate development app for your changes (web-viewer-test, desktop-viewer-test).
```

common/scripts/.eslintrc.ts.json
aruniverse: might be due to multiple version of typescript-eslint in the repo now due to using cra5 for the test-apps which pulls in ts-eslint@5, and eslint@8. i might try to fix this in a different pr

aruniverse (approved):

common/changes/@itwin/viewer-react/benp-optional-imodel-id_2022-09-06-18-35.json
aruniverse: lgtm!

PR #218, opened 2022-09-15
add `OptionalToUndefinedUnion` utility type to fix missing params from iTwinViewerInitializerParamSample and getIModelAppOptions

aruniverse:
- add `OptionalToUndefinedUnion` utility type to fix missing params from iTwinViewerInitializerParamSample and getIModelAppOptions
- rm imodelbank related code
- update to cra5
- update audit


ben-polinsky (approved):
LGTM

packages/modules/viewer-react/src/utilities/index.ts
```diff
@@ -33,7 +33,11 @@ export const isFunction = (item: any) => {
  * @returns
  */
 export const isEqual = (a?: any, b?: any) => {
-  if (!a || !b) {
+  if (a === undefined && b === undefined) {
+    return true;
+  }
+  // Do not check !a or !b as value can be boolean type
+  if (a === undefined || b === undefined) {
```
ben-polinsky, line 40: Ah missed this. I _think_ it could be simplified:
`a ?? b` if you can check for null as well as undefined.

ben-polinsky: Actually it'd be `!(a ?? b)`, I think

ben-polinsky: Nah disregard

PR #221, opened 2022-11-09
Add missing `redux` dependencies and `imodels-client-management` peer dependency

GytisCepk:
## Changes:
- Add missing _redux_ and _react-redux_ dependencies (peer dependencies of _@itwin/appui-react_):
  - Without dependencies, 2 versions of _@itwin/appui-react_ are being installed and it caused problems when using `UiFramework` static class.

- Add _@itwin/imodels-client-management_ as peer dependency in _@itwin/viewer-react_ package:
  - _@itwin/imodels-client-management_ is used in [code](https://github.com/iTwin/viewer/blob/11440b4b62ada422ca105665aa2f7b36875a545e/packages/modules/viewer-react/src/services/BaseInitializer.ts#L26) but only specified as dev dependency.


packages/modules/web-viewer-react/package.json
```diff
@@ -22,6 +22,7 @@
     "rebuild": "npm run clean && npm run build"
   },
   "dependencies": {
+    "@itwin/imodels-client-management": "^1.0.1",
```
GerardasB, line 25: Should this be in `peerDependencies`?

packages/modules/desktop-viewer-react/package.json
```diff
@@ -23,6 +23,7 @@
     "rebuild": "npm run clean && npm run build"
   },
   "dependencies": {
+    "@itwin/imodels-client-management": "^1.0.1",
```
GerardasB, line 26: Should this be in `peerDependencies`?

GytisCepk: Moved under `peerDependencies`.

packages/modules/desktop-viewer-react/package.json
```diff
@@ -44,6 +44,7 @@
     "@itwin/core-telemetry": "^3.2.0",
     "@itwin/ecschema-metadata": "^3.2.0",
     "@itwin/electron-authorization": "^0.8.3",
+    "@itwin/imodels-client-management": "^1.0.1",
```
aruniverse, line 47: can we make this a direct dep in the desktop viewer?

packages/modules/viewer-react/package.json
```diff
@@ -88,6 +88,7 @@
     "@itwin/core-geometry": "^3.2.0",
     "@itwin/core-react": "^3.2.0",
     "@itwin/imodels-access-frontend": "^1.0.1",
+    "@itwin/imodels-client-management": "^1.0.1",
```
aruniverse, line 91: sigh, good catch. technically new peers are breaking, but no one should be directly using this pkg, (since its supposed to be consumed directly in desktop & web viewer), so i'm more ok adding this as a peer, but lets make sure this is a minor bump?

packages/modules/web-viewer-react/package.json
```diff
@@ -39,6 +39,7 @@
     "@itwin/core-quantity": "^3.2.0",
     "@itwin/core-react": "^3.2.0",
     "@itwin/core-telemetry": "^3.2.0",
+    "@itwin/imodels-client-management": "^1.0.1",
```
aruniverse, line 42: same thing about possibily making this a direct dep

aruniverse (commented):
depending on what we end up doing with imodels-clients-* in this pr, if its breaking i might suggest we hold off on that part and put that in another pr. and we just publish an update with the redux and audit changes.

there are imodels-clients-*@2.0 pkgs we should probably update to along with the min core ver so might want to group those together in our next major ver bump

packages/modules/desktop-viewer-react/package.json
GytisCepk: Moved it to `dependencies`. 

Ideally, I think we would want this in peer (like other _iTwin_ packages), but in this case it would be a breaking change.

packages/modules/viewer-react/package.json
GytisCepk: This probably could be considered a non-breaking change since _@itwin/imodels-client-management_ was always required just not specified.

> lets make sure this is a minor bump?

Not completely sure what you mean by this. There is no other breaking changes to packages in this PR, so I guess this could be even a patch bump.

packages/modules/web-viewer-react/package.json
GytisCepk: Moved it to `dependencies`.

GerardasB (approved):

ben-polinsky (approved):

aruniverse:
ive been going back and forth so much with this pr; i hate peer deps and statics sooo so much.

redux/react-redux changes look fine

re the imodels-* pkgs:
- adding missing peer dep in viewer-react is def right
- i wonder if in web & desktop the imodels-* pkgs should actually be listed as just `devDeps` 
	- we've been adding transitive peers as dev deps 
	- this would eliminate the templates from having to have a dep on the client pkgs as well
		- and users would only need to add the dep on their end if they are overriding our defateults


heres a partial patch of what i think the changes should probably look like
[viewer-imodels-pr.txt](https://github.com/iTwin/viewer/files/10005916/viewer-imodels-pr.txt)

lmk what your thoughts are

aruniverse:
> heres a partial patch of what i think the changes should probably look like [viewer-imodels-pr.txt](https://github.com/iTwin/viewer/files/10005916/viewer-imodels-pr.txt)

looks like in my patch, i removed the direct dep in web-viewer & desktop-viewer react and just kept the direct deps at the test app/template levels. so the opposite of what i said yesterday... idk what was going on when i was wiritng that out yesterday.... sorry

in the patch, the imodels-client-management copies what we do for imodels-access-frontend.

as mentioned in todays call someone does need to have a direct dep on the client pkg, either web/desktop-viewer or the test app. im questioning if the imodels-access-frontend should be a direct dep in the web/desktop as well if we make the client a direct dep. im going to approve this pr for now, but i think we need to take a better look at these deps when we update to the 2.x imdoels pkgs

aruniverse (approved):

PR #228, opened 2023-01-03
Fix `rush audit`

GytisCepk:
Since last PR new security vulnerabilities were discovered in used packages. This PR updates package versions or ignores security vulnerabilities.

Updated packages (see commit messages).

Ignored errors:
- GHSA-9c47-m6qq-7p4h (need to confirm that this doesn't affect viewer)

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -69,41 +69,41 @@ importers:
       webpack: 4.42.0
     dependencies:
       '@bentley/icons-generic-webfont': 1.0.34
-      '@itwin/appui-abstract': 3.3.4_@itwin+core-bentley@3.3.4
-      '@itwin/appui-layout-react': 3.3.4_fbad76f3515a2ac0e20d3379feff0834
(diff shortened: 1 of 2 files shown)


ben-polinsky (approved):

common/config/rush/pnpm-lock.yaml
```diff
@@ -69,41 +69,41 @@ importers:
       webpack: 4.42.0
     dependencies:
       '@bentley/icons-generic-webfont': 1.0.34
```
aruniverse, line 71: why was lockfile updated?

aruniverse (approved):

common/config/rush/pnpm-lock.yaml
GytisCepk: > Next time could just do `rush update --full` to update the lockfile.

Wanted to update as little as possible because if dependencies introduced a new bug, it would have been hard to find the cause after a massive dep update.

Also, [`rush audit` output](https://github.com/iTwin/viewer/actions/runs/3828265461/jobs/6513659549#step:4:763) before updating lockfile. Most errors were caused by outdated versions in lockfile.

PR #233, opened 2023-01-27
Update links in the template. Remove unnecessary copyright headers

calebmshafer:
The links are not pointing to the correct part of the tutorials.

The copyright headers on the template files should not be there as we do not need to copyright the template code that we're giving out to everyone to use.


aruniverse (approved):
Should the desktop template be updated now too or defer that?

calebmshafer:
> Should the desktop template be updated now too or defer that?

@aruniverse yes we probably should. I was using the web one yesterday so was fixing things as I found them. Would have to go through the desktop one as well.

PR #250, opened 2023-03-28
Viewer rerender on sibling component change and web-viewer-test history fix

Jake-Screen:
In the change from v3.2.2 to v3.3.0 of viewer-react , a bug with the viewer was introduced. When there was a sibling component to the viewer in an viewer application, updating the state of the sibling component would cause the viewer to fully rerender. This was found to be caused by a change to the dependencies of  a hook in the IModelLoader. 

The change that caused the bug was changing the dependencies of the getModelConnection from checking only the iTwinId, iModelId, changeSetId, and filePath to checking all of the the viewer's props. When the parent of both the viewer and its sibling would rerender due to a change in the sibling, items like ui providers would be created again. This changed the viewer's props, causing the hook to trigger and reset the imodel connection. 
 
- update dependencies in getModelConnection hook from IModelLoader.tsx
    - removed viewerProps from dependencies
    - added viewerProps.iTwinId, viewerProps.iModelId, viewerProps.changeSetId, viewerProps.filePath to dependencies
- updated deprecated method from `getLocalizedStringWithNamespace` to `getLocalizedString`

The other change was to the web-viewer-test. The history component was deprecated and replaced with the history component from "@remix-run/router".

### DIFF
--- a/common/changes/@itwin/viewer-react/jake-rerender-bug-fix_2023-03-28-19-38.json
+++ b/common/changes/@itwin/viewer-react/jake-rerender-bug-fix_2023-03-28-19-38.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/viewer-react",
+      "comment": "Viewer rerender on sibling component change and web-viewer-test history fix",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/viewer-react"
+}
--- a/common/changes/@itwin/viewer-react/jake-rerender-bug-fix_2023-03-28-19-38.json
+++ b/common/changes/@itwin/viewer-react/jake-rerender-bug-fix_2023-03-28-19-38.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/viewer-react",
-      "comment": "Viewer rerender on sibling component change and web-viewer-test history fix",
+      "comment": "Fix Viewer rerendering on sibling component change",
       "type": "patch"
     }
   ],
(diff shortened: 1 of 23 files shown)


packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -63,10 +63,10 @@ const IModelLoader = React.memo((viewerProps: ModelLoaderProps) => {
 
     if (!requiredConnectionProps) {
       throw new Error(
-        IModelApp.localization.getLocalizedStringWithNamespace(
+        IModelApp.localization.getLocalizedString([
```
aruniverse, line 66: whats with the change here??

aruniverse (commented):
missing rush change

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
Jake-Screen: deprecated, was throwing errors, ill add it to description

common/changes/@itwin/viewer-react/jake-rerender-bug-fix_2023-03-28-19-38.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/viewer-react",
+      "comment": "Viewer rerender on sibling component change and web-viewer-test history fix",
```
aruniverse, line 5:
```suggestion
      "comment": "Viewer rerender on sibling component change",
```

the test app has nothing to do with this pkg

common/changes/@itwin/viewer-react/jake-rerender-bug-fix_2023-03-28-19-38.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/viewer-react",
+      "comment": "Viewer rerender on sibling component change",
```
aruniverse, line 5:
```suggestion
      "comment": "Fix Viewer rerendering on sibling component change",
```

PR #252, opened 2023-03-30
update peers to allow for range of `^3.7 || ^4.0.0-dev.x`

aruniverse:
update peers to allow temporarily use of ^3.7 for now
update web template to use 3.7 for vcr testing


packages/modules/desktop-viewer-react/package.json
```diff
@@ -72,11 +72,11 @@
     "typescript": "~4.4.0"
   },
   "peerDependencies": {
-    "@itwin/core-bentley": "^4.0.0-dev.41",
-    "@itwin/core-common": "^4.0.0-dev.41",
-    "@itwin/core-electron": "^4.0.0-dev.41",
-    "@itwin/core-frontend": "^4.0.0-dev.41",
-    "@itwin/ecschema-metadata": "^4.0.0-dev.41",
+    "@itwin/core-bentley": "^3.7.0 || ^4.0.0-dev.61",
```
raplemie, line 75: Weren't you the one to suggest `>=3.7.0` to address this ?

aruniverse: the viewer will use ^4.0.0 on core release, viewer will not use/support core 3.7 in the next minor.
this is just temp so we can get a viewer published we can use for testing

raplemie: Yep, but having the other scheme will give you less to update each time you want to :)

aruniverse: appui is using >= 3.7 because appui will support both 3.7 and 4. which is not something the viewer will do

aruniverse: > Yep, but having the other scheme will give you less to update each time you want to :)

true but doesnt hurt here too much

ben-polinsky (approved):

PR #258, opened 2023-04-17
rpc registration changes for desktop

aruniverse:
react to core-breaking changes wrt electron rpc registration


ben-polinsky (approved):

aruniverse:
@ben-polinsky , what do you think about the following breaking change:

Right now we have `viewerRpcs` in packages/apps/desktop-viewer-test/src/common/ViewerConfig.ts, which is in the `common` dir so it can be shared by both frontend and backend. Thoughts on:
- renaming the prop on the desktop viewer side from `additionalRpcInterfaces` to just `rpcInterfaces`?
- would we remove all default rpc registration in desktop-viewer-react, and move that up to the app level?
  - in our default template and test app, we would just import `viewerRpc` from common and pass it into the useDesktopInitializer hook and Viewer component

ben-polinsky (approved):

PR #259, opened 2023-04-19
Add `USING_NPM` env var in templates and default to true. Will be the case if bootstrapped using npx

aruniverse:
update template to always default to using npm, which will fix missing static assets


ben-polinsky:
Maybe a quick comment in the readme if not using NPM?

ben-polinsky (approved):

PR #264, opened 2023-04-27
Update css for Desktop Project Selection 

maggiemah:
Original: 

Edit 1:

Edit 2:

Edit 3:

### DIFF
--- a/packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.scss
+++ b/packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.scss
@@ -18,6 +18,8 @@
     .inline-input-with-button {
       display: flex;
       flex-direction: row;
+      margin-bottom: 5px;
+      width: 95%;
 
       .iui-input-container.iui-inline > .iui-input:last-child {
         min-width: 300px;
@@ -25,10 +27,26 @@
       }
 
       .iui-input-container {
-        grid-template: "label inputs icon" / auto minmax(50px, 290px) auto;
+        display: flex;
+        flex-direction: row;
+        width: 100%;
+        > .iui-label {
+          display: flex;
+          margin-left: 10px;
+          margin-right: 10px;
+          flex-wrap: wrap;
+          align-content: center;  
+        }
+        > .iui-input {
+          border: 2px solid var(--iui-text-color-muted);
+          background-color: var(--iui-color-background-1);
+          width: 100%;
+          padding: 10px;
+        }
       }
 
       .iui-button {
+        background: 0;
         border-top-left-radius: 0;
         border-bottom-left-radius: 0;
         margin-left: -50px;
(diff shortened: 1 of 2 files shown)


packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.scss
```diff
@@ -18,17 +18,35 @@
...
       }
 
       .iui-input-container {
-        grid-template: "label inputs icon" / auto minmax(50px, 290px) auto;
+        display: flex;
+        flex-direction: row;
+        width: 100%;
+        > .iui-label {
+          display: flex;
+          margin-left: 10px;
```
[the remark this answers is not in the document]
maggiemah: I can use margin shorthand syntax,

aruniverse:
Take a look at: https://itwindemo.bentley.com/view
I think its supposed to look like this, without the create project button

Also, make sure you update the same file in the template/desktop-viewer

ben-polinsky:
Looks solid, @aruniverse not sure if you want the search on the right, in which case should be a simple fix.

aruniverse:
> Looks solid, @aruniverse not sure if you want the search on the right, in which case should be a simple fix.

Yea lets, do the search on the right

@maggiemah

packages/apps/desktop-viewer-test/src/frontend/components/modelSelector/SelectProject.scss
```diff
@@ -18,20 +18,72 @@
...
         min-width: 300px;
         padding-right: 61px;
       }
 
       .iui-input-container {
-        grid-template: "label inputs icon" / auto minmax(50px, 290px) auto;
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+        font-family: "Open Sans",-apple-system,BlinkMacSystemFont,"Segoe UI","Roboto","Helvetica Neue",Arial,sans-serif;
```
aruniverse, line 33: wait, do you really need to set all this extra stuff? 
is this diverging too much from the design standards made by the components?

maggiemah: Good point, defining the family isn't necessary but explicitly inheriting font-family and font-size is needed to change iui-input and iui-tab.

aruniverse: any reason the font family or size needs to be changed at all?

maggiemah: Seems to match better

font-family + font-size Inherited: 
image

Without: 
image

aruniverse: ah dont worry too much about the exact look, the link i pointed too is using like a 2 yr old itiwnui ver
i wouldnt try to override the look too much, defeats the purpose of having our own custom library / design system

PR #276, opened 2023-05-24
Finalize release

aruniverse:


packages/apps/desktop-viewer-test/src/frontend/components/home/Home.scss
ben-polinsky, line 1: How are css variables pulled in from itwinui?

packages/templates/cra-template-desktop-viewer/template/src/frontend/components/modelSelector/BriefcaseStatus.scss
```diff
@@ -4,6 +4,7 @@
 *--------------------------------------------------------------------------------------------*/
 @import '@itwin/itwinui-variables/index.css';
```
ben-polinsky, line 5: Why does this import remain but in other scss files it is removed?

packages/apps/desktop-viewer-test/src/frontend/components/home/Home.scss
aruniverse: theyre pulled in via the global css file being imported at the root

packages/templates/cra-template-desktop-viewer/template/src/frontend/components/modelSelector/BriefcaseStatus.scss
aruniverse: thank you, nice catch

PR #290, opened 2023-08-01
[templates] consume tree-widget / property-grid 1.0.0

aruniverse:


packages/apps/desktop-viewer-test/package.json
ben-polinsky, line 1: gh took the lazy way out on this diff

ben-polinsky: Ah there's a toggle :)

ben-polinsky (approved):

PR #295, opened 2023-08-25
Remove auth check when viewing BlankConnections 

AnDuong249:
- Currently, blank connections still require users to authenticate before rendering the Viewer so I'm adding a BlankConnection check to render the Viewer even if the user doesn't have accessToken.
- Also removing the iTwinId requirements when creating a BlankConnection Viewer so the users don't have to pass in the iTwinId props when they're creating one.


ben-polinsky (approved):

ben-polinsky:
Would like @aruniverse's thoughts before merging.

packages/modules/viewer-react/src/components/BaseViewer.tsx
```diff
@@ -29,9 +29,10 @@ export const BaseViewer = ({
   });
 
   const accessToken = useAccessToken();
+  const isBlankConnection = loaderProps.extents && loaderProps.location;
```
aruniverse, line 32: if its a blank connection and an iTwinId is provided we do want an authclient; the itwinid is used to fetch & display reality data, which is often the most common use case for BlankConnections


suggested reading:
- https://www.itwinjs.org/learning/frontend/blankconnection/
- https://www.itwinjs.org/reference/core-frontend/imodelconnection/blankconnection/
- https://www.itwinjs.org/reference/core-frontend/imodelconnection/blankconnectionprops/

packages/modules/viewer-react/src/components/BaseViewer.tsx
```diff
@@ -29,9 +29,10 @@ export const BaseViewer = ({
   });
 
   const accessToken = useAccessToken();
+  const isBlankConnection = loaderProps.extents && loaderProps.location && !loaderProps.iTwinId;
```
aruniverse, line 32: so this will work, but wondering if theres a better way

how would users know when they supply an itwinid that they need to provide auth?
can there be some type magic done for this?

aruniverse: mind updating the readme while youre at this?

packages/modules/web-viewer-react/src/services/Initializer.ts
```diff
@@ -36,14 +37,13 @@ export class WebInitializer {
   };
 
   /** Web viewer startup */
-  public static async startWebViewer(options: WebInitializerParams) {
+  public static async startWebViewer(options: WebInitializerParams, authClient?: ViewerAuthorizationClient) {
```
aruniverse, line 40: could you just do a union for the type here to not add an extra arg?

ben-polinsky: I think this should go into the `WebInitializerParams` type in types.ts?

packages/modules/web-viewer-react/src/services/Initializer.ts
aruniverse, line 1: revert changes to this file

packages/modules/desktop-viewer-react/src/components/Viewer.tsx
```diff
@@ -10,6 +10,6 @@ import { useDesktopViewerInitializer } from "../hooks";
 import type { DesktopViewerProps } from "../types";
 
 export const Viewer = (props: DesktopViewerProps) => {
-  const initialized = useDesktopViewerInitializer(props);
+  const initialized = useDesktopViewerInitializer({...props, clientId: ""});
```
ben-polinsky, line 13: Won't this override the `clientId` to always be blank? Desktop apps still need to provide a `clientId` if they're not using a local snapshot/briefcase.

packages/modules/desktop-viewer-react/src/types.ts
```diff
@@ -22,7 +22,7 @@ export type DesktopViewerProps = XOR<
   XOR<FileViewerProps, BlankViewerProps>,
   ConnectedViewerProps
 > &
-  DesktopInitializerParams;
+  Omit<DesktopInitializerParams, "clientId">;
```
ben-polinsky, line 25: Continuing off the last comment, not sure why we'd want to omit the clientId as a prop. Users will still need to provide this, when necessary.

ben-polinsky (approved):

PR #297, opened 2023-09-05
Sync Selection Scope List, Active Selection Scope & Selection Count between Presentation and AppUI

AnDuong249:
- The selection count in the status bar counting all the elements an assembly is composed of instead of treating the assembly as a single entity when the selection scope is assembly / top assembly
- The underlying cause of this issue is that AppUi 4.x severed its dependencies on Presentation and instead of syncing from the Presentation.selection.selectionChanged. Does not syncs with the iModel.selectionSet.onChanged which won't consider the selection scope when counting the selection
- Solution proposed in the AppUI issue https://github.com/iTwin/appui/pull/323 was to split off a version of that element selection count status bar component that doesn't force it to be synced with the redux store.
- The change in this PR includes a workaround that overrides the redux store state for the selected element count by adding a listener to the iModelConnection.selectionSet.onChanged from the handler of the Presentation.selection.onChanged event to sync it back with that one
- Long term presentation would expose a hook or hoc to wrap that field with to have it easily sync with their selection change event

Reference: https://dev.azure.com/bentleycs/beconnect/_git/TCDEAppService/pullrequest/338295
- Added syncing active selection scope when first called since when initialized, the Presentation active scope is undefined.

### DIFF
--- a/common/changes/@itwin/viewer-react/SyncSelectionScope_2023-09-05-17-30.json
+++ b/common/changes/@itwin/viewer-react/SyncSelectionScope_2023-09-05-17-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/viewer-react",
+      "comment": "Sync Selection Scope List, Active Selection Scope & Selection Count between Presentation and AppUI",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/viewer-react"
+}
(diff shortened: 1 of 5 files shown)


hl662 (approved):
LGTM

aruniverse (approved):

Ben-Pusey-Bentley (approved):

PR #311, opened 2024-02-23
Updated supported node version range in rush

hl662:
- Bumps up supported node version in rush, matching that in itwinjs core. Also update rush and pnpm version to support node 20
- Ran rush update, which auto modifies some common/scripts files that rush uses
- Update `webpack` version in `desktop-viewer-test` to address high security vulnerability

Tested CI workflow and ran desktop viewer test using latest version of node 20

### DIFF
--- a/packages/apps/desktop-viewer-test/package.json
+++ b/packages/apps/desktop-viewer-test/package.json
@@ -97,7 +97,7 @@
     "rimraf": "^3.0.2",
     "sass": "^1.64.2",
     "typescript": "~5.0.4",
-    "webpack": "4.42.0"
+    "webpack": "^5.1.2"
   },
   "eslintConfig": {
     "extends": [
--- a/rush.json
+++ b/rush.json
@@ -1,11 +1,11 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/rush.schema.json",
-  "rushVersion": "5.100.1",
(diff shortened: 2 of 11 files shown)


packages/apps/desktop-viewer-test/package.json
```diff
@@ -92,12 +92,11 @@
     "cpx2": "4.2.0",
     "cross-env": "^5.2.1",
     "electron-devtools-installer": "^2.2.4",
-    "eslint": "^7.32.0",
     "npm-run-all": "^4.1.5",
     "rimraf": "^3.0.2",
     "sass": "^1.64.2",
     "typescript": "~5.0.4",
-    "webpack": "4.42.0"
+    "webpack": "^5.1.2"
```
aruniverse, line 99: why did we jump majors here?
did we react to the breaking changes? this "instance" of webpack is being used to bundle the backend

rush.json
```diff
@@ -1,11 +1,11 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/rush.schema.json",
-  "rushVersion": "5.100.1",
-  "pnpmVersion": "6.24.4",
+  "rushVersion": "5.112.2",
+  "pnpmVersion": "7.32.2",
```
aruniverse, line 4: make sure we update the use of pnpm in the pipelines too

aruniverse (approved):

PR #312, opened 2024-02-26
Add IModelApp.shutdown on page unload events

hl662:
To address a memory leak issue caused by frequent page refreshes, an explicit IModelApp.shutddown() was added inside of a 'beforeunload' event listener. Page refreshes, as well as navigation away from a react component, do not count as a componentWillUnmount() react lifecycle, instead it is a part of a 'beforeunload' window DOM event.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -74,69 +74,69 @@ importers:
       webpack: 4.42.0
     dependencies:
       '@bentley/icons-generic-webfont': 1.0.34
-      '@itwin/appui-abstract': 4.2.4_@itwin+core-bentley@4.2.4
-      '@itwin/appui-layout-react': 4.7.0_d986596b56cdae44ca6c9dea6b97a471
-      '@itwin/appui-react': 4.7.0_4120ccd471dca5d63d12e01f8bb9b3d6
-      '@itwin/components-react': 4.7.0_590a8bc26b7a158d306e1a2edd1200fc
-      '@itwin/core-backend': 4.2.4_eef9b76cf0304df1d1863ac831804c5c
-      '@itwin/core-bentley': 4.2.4
-      '@itwin/core-common': 4.2.4_c7c35732681224dfce00cc4f83585540
-      '@itwin/core-electron': 4.2.4_3811dd19316a5c2ec37c95f2ce86a990
-      '@itwin/core-frontend': 4.2.4_68ed6f323248b29ff2df1a8fc647061a
-      '@itwin/core-geometry': 4.2.4
-      '@itwin/core-i18n': 4.2.4_@itwin+core-bentley@4.2.4
-      '@itwin/core-markup': 4.2.4_42ad66ec6c13f645f964574f5c14e992
-      '@itwin/core-orbitgt': 4.2.4
-      '@itwin/core-quantity': 4.2.4_@itwin+core-bentley@4.2.4
-      '@itwin/core-react': 4.7.0_3c550e1bb2007176651b8dc902f0dcb9
-      '@itwin/core-telemetry': 4.2.4_@itwin+core-geometry@4.2.4
+      '@itwin/appui-abstract': 4.4.0_@itwin+core-bentley@4.4.0
+      '@itwin/appui-layout-react': 4.8.3_59726ffc379f36970c5d2c060a633f87
+      '@itwin/appui-react': 4.9.0_1464ab937f5f23dc241ba346c0a57550
+      '@itwin/components-react': 4.9.0_3ffd1b6506838f2d12fbe6ffbba27a2f
+      '@itwin/core-backend': 4.4.0_84beed194d57ae4df09620ef7918dae1
+      '@itwin/core-bentley': 4.4.0
+      '@itwin/core-common': 4.4.0_15e4dedc8df929df1241423c58deb905
+      '@itwin/core-electron': 4.4.0_3dffe86b29b6bf6fa658b7721185eade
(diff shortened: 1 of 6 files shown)


packages/modules/web-viewer-react/src/components/Viewer.tsx
```diff
@@ -3,13 +3,25 @@
...
+import React, { useEffect } from "react";
 
 import { useWebViewerInitializer } from "../hooks";
 import type { WebViewerProps } from "../types";
 
 export const Viewer = (props: WebViewerProps) => {
   const initialized = useWebViewerInitializer(props);
+  useEffect(() => {
+    const handleBeforeUnload = async () => {
+      IModelApp.shutdown()
```
aruniverse, line 17: take a look at `WebInitializer.cancel`

also wonder if we need to manually dispose of other stuff ourselves like presentation, but wonder if they listen to an event on imodelapp shutdown

hl662: Other stuff would have to register a [onbeforeShutdown](https://github.com/iTwin/itwinjs-core/blob/3cc2df33f158c51d02fb17a3c3d4d303c857acd8/core/frontend/src/IModelApp.ts#L442) event, and it seems to imply that they would have to manually be disposed. Is this event handled by us, or the team in charge of that particular stuff, I'm not sure.

If we pass IModelApp.shutdown() into WebInit.cancel(), we would have to be careful of the side-effects of `.shutdown()`. Because inherently, we are looking to free up resources, and a question we would have is 'how many resources do we really need to free up` to prevent this memory leak? We know that `_renderSystem` needs to be disposed, but what about the other things disposed in shutdown(), like the authClient, entityClasses, toolAdmin, viewManager, and tileAdmin?

hl662: i looked at WebInitializer.cancel and init(), two things popped up to me - IModelApp does have a [extensionAdmin](https://github.com/iTwin/itwinjs-core/blob/74610de97a26fc395842b34adf63dbfd7c9432e4/core/frontend/src/extension/ExtensionAdmin.ts#L44), which the web viewer calls its' registerHost() on init if an extension is registered. IModelApp.shutdown() doesn't do anything dispose wise for extensionAdmin, do you think we need to do anything there?

Only other thing would be [ViewerAuthorization.client](https://github.com/iTwin/viewer/blob/b4d784da8456f8043548c004b6d8506e20cacaf4/packages/modules/web-viewer-react/src/services/Initializer.ts#L59), it will hold a reference to an authClient, that wouldn't be cleared up, but I don't think this is that big of an issue?

aruniverse: Theres nothing apart of the extension admin that i think needs to be disposed
It should be ok to just set IModelApp.authorizationClient = undefined, we dont want to dispose of the actual client

common/config/rush/pnpm-lock.yaml
```diff
@@ -73,70 +73,70 @@ importers:
...
-      '@itwin/core-telemetry': 4.2.4_@itwin+core-geometry@4.2.4
+      '@itwin/appui-abstract': 4.4.0_@itwin+core-bentley@4.4.0
+      '@itwin/appui-layout-react': 4.8.3_lfzg77bxt43jodc5fqdauyz7q4
+      '@itwin/appui-react': 4.9.0_crskxe37l4r5yja3undmbjlvka
+      '@itwin/components-react': 4.9.0_h76rwzigqohs2ex34373xit2f4
+      '@itwin/core-backend': 4.4.0_qs7o2gknk6xe34ewedxxsgg24e
+      '@itwin/core-bentley': 4.4.0
+      '@itwin/core-common': 4.4.0_cxsn5xen7eu56esbii6frxvzau
+      '@itwin/core-electron': 4.4.0_hx76q2zjw27w7jsyw5zbdbpk3y
+      '@itwin/core-frontend': 4.4.0_r2oyeiz6cvkfaphvsx2dlem3ea
```
aruniverse, line 84: can you update lockfile to also pull in the 4.4.4 patch which you had a patch for dropping vps ?

packages/modules/web-viewer-react/src/components/Viewer.tsx
aruniverse: does appui need to be shut down?

hl662: no, appui doesn't - the root of the problem lies in any possible webgl textures not being disposed of correctly, and that's not something app ui is responsible for. While app ui does have a [UiAdmin](https://github.com/iTwin/itwinjs-core/blob/ec2d5b0b84d08b2dda6bf3a8765bb2feb64ac777/ui/appui-abstract/src/appui-abstract/UiAdmin.ts#L245) class, it does not contain a shutdown() that other admin classes have.

common/config/rush/pnpm-lock.yaml
hl662: I could have sworn I ran `rush purge` then re ran `rush update` to bump up the lockfile, weird. I deleted pnpm-lock file manually and then ran rush update to make sure, and the 4.4.4 patch is in now.

aruniverse: should avoid manually deleting lock file

use `rush update --full` next time

aruniverse (approved):

hl662:
Added to desktop-viewer-react, tested using desktop viewer-test with a local .bim file (electron lets you do window reloads). Same behavior as the browser version so no differences with the implementation in web-viewer-react

PR #314, opened 2024-05-23
Add ability to initialize Presentation with custom props

saskliutas:
Added ability to supply custom props to use during Presentation initialization. This is need to allow supplying new unified selection storage to be used by presentation-frontend.

Remove code setting active locale as it is done by default when initializing Presentation: https://github.com/iTwin/itwinjs-core/blob/73187a864fe12a1ca44d38f799bc9a2d79dfa40a/presentation/frontend/src/presentation-frontend/Presentation.ts#L74

### DIFF
--- a/packages/modules/viewer-react/src/services/BaseInitializer.ts
+++ b/packages/modules/viewer-react/src/services/BaseInitializer.ts
@@ -191,11 +191,7 @@ export class BaseInitializer {
       yield UiFramework.initialize(undefined);
 
       // initialize Presentation
-      yield Presentation.initialize({
-        presentation: {
-          activeLocale: IModelApp.localization.getLanguageList()[0],
-        },
-      });
+      yield Presentation.initialize(viewerOptions?.presentationProps);
 
       // Sync selection count & active selection scope between Presentation and AppUi. Runs after the Presentation is initialized.
       syncSelectionCount();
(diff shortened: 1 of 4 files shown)


ben-polinsky:
Looks good to me. We'll need to mitigate this audit failure: 

ERROR: HIGH Security Vulnerability: Path traversal in webpack-dev-middleware in webpack-dev-middleware (from ..__..__packages__apps__desktop-viewer-test>@bentley/react-scripts>webpack-dev-server>webpack-dev-middleware).  See https://github.com/advisories/GHSA-wr3j-pwj9-hqq6 for more info.

ben-polinsky (approved):

packages/modules/viewer-react/src/services/BaseInitializer.ts
```diff
@@ -191,11 +191,7 @@ export class BaseInitializer {
       yield UiFramework.initialize(undefined);
 
       // initialize Presentation
-      yield Presentation.initialize({
-        presentation: {
-          activeLocale: IModelApp.localization.getLanguageList()[0],
```
aruniverse, line 196: do we not need to provide a default for the active locale?

aruniverse (approved):

packages/modules/viewer-react/src/services/BaseInitializer.ts
saskliutas: It is optional and Presentation.initialize does the same what viewer was doing https://github.com/iTwin/itwinjs-core/blob/73187a864fe12a1ca44d38f799bc9a2d79dfa40a/presentation/frontend/src/presentation-frontend/Presentation.ts#L74

Issue #317, opened 2024-06-20
Integrate `@itwin/unified-selection` into viewer


saskliutas:
Integrate `@itwin/unified-selection` for unified selection handling. If `selectionStorage` is passed to `Viewer` component it should use [`enableUnifiedSelectionSyncWithIModel`](https://github.com/iTwin/presentation/tree/master/packages/unified-selection#imodel-selection-synchronization-with-unified-selection) to synchronize hilited element with unified selection instead of [viewWithUnifiedSelection](https://github.com/iTwin/viewer/blob/0359ac61df1dab0c330b337ca77f1a6e87ac4fd3/packages/modules/viewer-react/src/components/app-ui/providers/UnifiedSelectionViewportControl.tsx#L17) HOC

aruniverse:
Will this be a direct dep or a new peer?

PR #319, opened 2024-07-01
Integrate unified selection

jasdom:
Closes https://github.com/iTwin/viewer/issues/317.


packages/modules/viewer-react/package.json
```diff
@@ -33,31 +33,34 @@
     "@itwin/itwinui-illustrations-react": "^2.0.1",
     "@itwin/itwinui-react": "^2.6.0",
     "@itwin/reality-data-client": "^1.0.0",
+    "@itwin/unified-selection": "^0.4.4",
     "lodash.isequal": "^4.5.0"
   },
   "devDependencies": {
-    "@itwin/appui-abstract": "^4.0.0",
-    "@itwin/appui-layout-react": "^4.0.0",
-    "@itwin/appui-react": "^4.0.0",
```
aruniverse, line 41: I know appui has killed off appui-layout-react, but since we still techniclaly list it as a peer dep, should we be removing the dev dep on it?

aruniverse: And this propagates through the other viewer pkgs

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
```diff
@@ -55,7 +59,7 @@ export class DefaultContentGroupProvider extends ContentGroupProvider {
       contents: [
         {
           id: "iTwinViewer.UnifiedSelectionViewport",
-          classId: UnifiedSelectionViewportControl,
+          classId: this._syncWithUnifiedSelectionStorage ? IModelViewportControl : UnifiedSelectionViewportControl,
```
aruniverse, line 62: is the order here correct? should it be reveresed?

packages/modules/viewer-react/package.json
```diff
@@ -33,31 +33,34 @@
...
+    "@itwin/core-frontend": "^4.7.3",
+    "@itwin/core-geometry": "^4.7.3",
+    "@itwin/core-i18n": "^4.7.3",
+    "@itwin/core-markup": "^4.7.3",
+    "@itwin/core-orbitgt": "^4.7.3",
+    "@itwin/core-quantity": "^4.7.3",
+    "@itwin/core-react": "4.15.0",
+    "@itwin/core-telemetry": "^4.7.3",
+    "@itwin/ecschema-metadata": "^4.7.3",
+    "@itwin/ecschema-rpcinterface-common": "^4.7.3",
```
aruniverse, line 55: are we really able to only add these new deps as just dev deps? they dont need to be peers?

aruniverse (commented):
Wonder if we should defer this change to when we do the next major version

packages/modules/viewer-react/package.json
saskliutas: Nice catch. It looks like `"@itwin/ecschema-rpcinterface-common"` should be peer dep. Although I think it would be nice to avoid it. Maybe instead of creating `SchemaContext` in viewer we could take a prop like `getSchemaContext?: (imodel: IModelConnectio) => SchemaContext`. This would allow to reuse same SchemaContext between Viewer and other components. 

Also these deps should be moved under `dependencies`:
```
"@itwin/presentation-core-interop": "^0.2.3",
"@itwin/presentation-shared": "^0.3.1",
```

jasdom: Added a `getSchemaContext` prop instead of creating `SchemaContext` and moved the dependencies.

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.ts
jasdom: Yes, it is correct. Before, we used the `viewWithUnifiedSelection` HOC to provide synchronization with unified selection; however, with these changes, the `enableUnifiedSelectionSyncWithIModel` hook is used instead, and so we no longer need to wrap `IModelViewportControl`.

packages/modules/viewer-react/package.json
jasdom: Right, reverted the changes to appui dependencies.

packages/modules/viewer-react/src/types.ts
```diff
@@ -33,6 +33,7 @@ import type {
   XYAndZ,
 } from "@itwin/core-geometry";
 import { PresentationProps } from "@itwin/presentation-frontend";
```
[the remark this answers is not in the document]
jasdom: Changed.

packages/modules/viewer-react/src/types.ts
```diff
@@ -33,6 +33,7 @@ import type {
...
+import { SelectionStorage } from "@itwin/unified-selection";
```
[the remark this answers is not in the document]
jasdom: Changed.

grigasp:
> Wonder if we should defer this change to when we do the next major version

Hey @aruniverse. Can you explain the reason? The PR has no breaking changes, as far as I can tell.

packages/modules/viewer-react/src/hooks/useUnifiedSelectionSync.tsx
```diff
@@ -0,0 +1,80 @@
...
+        ...createECSqlQueryExecutor(iModelConnection),
+        ...createCachingECClassHierarchyInspector({ schemaProvider: createECSchemaProvider(schemaContext) }),
+        key: iModelConnection.key,
+        hiliteSet: iModelConnection.hilited,
+        selectionSet: iModelConnection.selectionSet,
+      },
+      selectionStorage,
+      activeScopeProvider: getActiveScope,
+    });
+  }, [iModelConnection, selectionStorage]);
```
aruniverse, line 39: is the dep array here missing `getSchemaContext`?

packages/modules/viewer-react/package.json
```diff
@@ -30,35 +30,38 @@
...
+    "@itwin/presentation-shared": "^0.3.1",
     "@itwin/itwinui-illustrations-react": "^2.0.1",
     "@itwin/itwinui-react": "^2.6.0",
     "@itwin/reality-data-client": "^1.0.0",
+    "@itwin/unified-selection": "^0.4.5",
     "lodash.isequal": "^4.5.0"
   },
   "devDependencies": {
-    "@itwin/appui-abstract": "^4.0.0",
+    "@itwin/appui-abstract": "^4.7.3",
```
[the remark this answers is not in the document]
aruniverse: ah is this updated becasue of rush complaining about inconsistent versions?

aruniverse (approved):

aruniverse:
> > Wonder if we should defer this change to when we do the next major version
> 
> Hey @aruniverse. Can you explain the reason? The PR has no breaking changes, as far as I can tell.

yes no more breaking changes in the pr. i was worried about the original changes, but saulius' [suggestion here](https://github.com/iTwin/viewer/pull/319#discussion_r1670703861) helps us avoid a major

common/changes/@itwin/desktop-viewer-react/integrate-unified-selection_2024-07-02-08-51.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/desktop-viewer-react",
+      "comment": "Integrated `@itwin/unified-selection` for unified selection handling when `SelectionStorage` is provided.",
```
aruniverse, line 5: actually think you can set the comment to an empty string, and type to none.
you made the change in the viewer, and that should propagate up into these pkgs automatically

common/changes/@itwin/web-viewer-react/integrate-unified-selection_2024-07-02-08-51.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/web-viewer-react",
+      "comment": "Integrated `@itwin/unified-selection` for unified selection handling when `SelectionStorage` is provided.",
```
aruniverse, line 5: same comment as desktop viewer

packages/modules/viewer-react/package.json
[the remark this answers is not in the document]
grigasp: > ah is this updated becasue of rush complaining about inconsistent versions?

yes, I believe so

packages/modules/viewer-react/src/hooks/useUnifiedSelectionSync.tsx
grigasp: yes, added

packages/modules/viewer-react/src/types.ts
```diff
@@ -102,6 +104,10 @@ export interface LoaderProps {
   viewCreatorOptions?: ViewerViewCreator3dOptions;
   /** Component to show when loading iModel key */
   loadingComponent?: React.ReactNode;
+  /** unified selection storage to synchronize with. Should be passed together with `getSchemaContext` */
```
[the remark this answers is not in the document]
grigasp: Yes and no... 

The `useUnifiedSelectionSync` makes sure that `IModelConnection` selection changes are synchronized with the new unified selection storage; the hook is used by `IModelLoader`, which means we need those two props on `LoaderProps`.

However, we also need `Presentation` library to use the exact same storage to make sure that components using `SelectionManager` from `@itwin/presentation-frontend` get notified about changes in the storage and can change its content. This is done by passing the storage to `Presentation.initialize` call, which is done by the `BaseInitializer.initialize`, which takes `ViewerInitializerParams`. So we need at least `selectionStorage` prop on `ViewerInitializerParams`.

Here's what I suggest:
- Revert the current change and move the two props to a new interface `UnifiedSelectionProps`.
- In `ViewerInitializerParams` keep only `selectionStorage?: UnifiedSelectionProps["selectionStorage"]`. I think this could be avoided, because `useBaseViewerInitializer` receives selection storage and could inject it into `PresentationProps` that are passed to `BaseInitializer.initialize`. However, I have a hard time understanding what's happening in `useBaseViewerInitializer` to make any changes there confidently.
- Extend `ModelLoaderProps` with `& UnifiedSelectionProps`.
- Change `ViewerCommonProps` to `Omit<ViewerInitializerParams, "selectionStorage"> & LoaderProps & UnifiedSelectionProps`.

common/changes/@itwin/desktop-viewer-react/integrate-unified-selection_2024-07-02-08-51.json
grigasp: I'm not sure I agree. `@itwin/viewer-react` is a direct dependency of `@itwin/desktop-viewer-react` and consumers are expected to consume `@itwin/desktop-viewer-react` directly, aren't they? In that case, the change in `@itwin/viewer-react` will be hidden from them, even though the API additions will become available in `@itwin/desktop-viewer-react`.

aruniverse: True, the hope has always been users look at the changelog for viewer-react itself, but I see your point about it getting lost in the weeds. This is a minor nit so resolving

packages/modules/viewer-react/src/types.ts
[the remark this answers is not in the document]
aruniverse: I think your suggestions sound good, I'll review the useBaseViewerInitializer changes you're proposing, like you said not sure if needed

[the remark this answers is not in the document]
grigasp: I think I managed to supply the required props without introducing anything new to `ViewerInitializerParams`. Please have a look.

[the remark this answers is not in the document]
aruniverse: lgtm, thx

aruniverse (approved):

PR #322, opened 2024-07-11
Integrate Components

ben-polinsky:
Allows viewing _Components_ by setting a the `contextId`, `componentId`, and `documentId` for a corresponding component in component center. 

Only works with backends which support opening component models.


packages/modules/viewer-react/src/tests/components/iModel/IModelLoader.test.tsx
```diff
@@ -126,6 +126,8 @@ jest.mock("@itwin/core-frontend", () => {
     SheetViewState: {
       className: "",
     },
+    FrontendLoggerCategory: {},
```
ben-polinsky, line 129: These mocks are getting out of control and make our tests feeble. While I know this Viewer is on the back burner, it does seem like it's still being used as an initial testing ground for new features for our primary viewer. We may want to take a look at this.

ben-polinsky:
Turning to a draft as this most likely will never make it into the viewer. For reference.

packages/modules/viewer-react/src/tests/components/iModel/IModelLoader.test.tsx
hl662: We could look into switching to vitest?

ben-polinsky: I don't think the issue is with the library - it's how we use it. You can mock the world with any test library, that doesn't mean it's a good idea.

ben-polinsky: But yeah I like vitest :)

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
```diff
@@ -83,20 +101,31 @@ const ViewerHome: React.FC = () => {
...
+  }, []);
+
+  useEffect(() => {
+    let url = `viewer?`;
+    allParams.forEach((param) => {
+      const sep = url.endsWith("?") ? "" : "&";
+      if (param) {
+        url += `${sep}${param}`;
+      }
+    });
```
tcobbs-bentley, line 125: I think this section can be replaced with the following:

```ts
    const url = `viewer?${allParams.join("&")`
```

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
```diff
@@ -125,16 +154,19 @@ const ViewerHome: React.FC = () => {
...
 
   return (
     <div style={{ height: "100vh" }}>
       <Viewer
         authClient={authClient}
-        iTwinId={iTwinId ?? ""}
-        iModelId={iModelId ?? ""}
-        changeSetId={changesetId}
+        iTwinId={iTwinId}
+        iModelId={iModelId}
```
tcobbs-bentley, line 165: Unlike iTwinId, iModelId hasn't been updated to default to empty string. It could be that the old code was broken, and it could be that other changes you made make it so undefined is the desired value here, but it could also be that this change is a mistake.

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
tcobbs-bentley: I see now that my suggested change missed the `if (param)` line, although `filter` could be used to rectify that.

However, I don't understand how this new URL can work. It is lacking the `<name>=` part for each param. So, while before it would be something like `viewer?iTwinId=<blah>&iModelId=<yada>`, now you just have `viewer?<blah>&<yada>&&`.

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
```diff
@@ -125,16 +154,19 @@ const ViewerHome: React.FC = () => {
...
+        // component props:
```
tcobbs-bentley, line 166: Why was changeSetId removed?

packages/modules/viewer-react/src/services/BaseInitializer.ts
```diff
@@ -71,6 +71,7 @@ const syncActiveSelectionScope = () => {
   Presentation.selection.scopes.activeScope =
     UiFramework.getActiveSelectionScope();
 
+  // eslint-disable-next-line deprecation/deprecation
```
tcobbs-bentley, line 74: I think you can just remove the explicit type for `args` and it will be inferred, thus removing the need to use the deprecated `UiSyncEventArgs` type.

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
ben-polinsky: Yeah thanks. That's a miss on my part.

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
ben-polinsky: accident... fixed

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
ben-polinsky: I've set all to themselves or an empty string now, thanks.

packages/modules/viewer-react/src/services/BaseInitializer.ts
ben-polinsky: Yep. Thanks again. Fixed.

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
tcobbs-bentley: Is the trailing `&` at the end of the new url OK, or should it be removed after computing the url?

packages/modules/viewer-react/src/services/BaseInitializer.ts
tcobbs-bentley: You left the `UiSyncEventArgs` import at the top of the file.

packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx
ben-polinsky: Right, yeah I think it's okay.

packages/modules/viewer-react/src/services/iModel/ComponentConnection.ts
```diff
@@ -0,0 +1,260 @@
...
+ * - getRpcProps is overridden to provide "correct" props for the Component model rather than those returned from getConnectionProps.
+ *
+ * @See https://docs.bentley.com/LiveContent/web/AECOsim%20Building%20Designer%20Help-v2/en/GUID-CCDE2FF6-C63E-4876-B3AE-5F2B33EF8547.html
+ * for more about components.
+ */
+
+export class ComponentConnection extends IModelConnection {
+  public iModelProps: IModelConnectionProps;
+  private _documentId: string;
+  private _isClosed?: boolean = false;
```
tcobbs-bentley, line 50: Why is this an optional value? Shouldn't it just be:

```ts
  private _isClosed = false;
```

Note that doing this would also simplify `isClosed`.

packages/modules/viewer-react/src/services/iModel/ComponentConnection.ts
```diff
@@ -0,0 +1,260 @@
...
+        "IModelReadRpcInterface.getConnectionProps() is not available"
+      );
+    }
+    openForReadOperation.policy.retryInterval = () => connectionRetryInterval;
+
+    Logger.logTrace(loggerCategory, `IModelConnection.open`, iModelProps);
+    const startTime = Date.now();
+
+    const removeListener = RpcRequest.events.addListener(
+      // eslint-disable-next-line deprecation/deprecation
```
tcobbs-bentley, line 132: The explicit types for type and request probably aren't needed, which also makes this eslint-disable comment unneeded.  The following eslint-disable comment is obviously still needed due to using `RpcRequestEvent`.

PR #328, opened 2024-09-03
Fix desktop viewer unified imports

ben-polinsky:
Fix a couple of incorrect imports in the desktop viewer template.


aruniverse (approved):
can you check web template imports too?

ben-polinsky:
> can you check web template imports too?
I did an eye test, but I'll double check.

PR #329, opened 2024-09-09
register ECSchemaRpcInterface with web viewer template

ben-polinsky:
Web viewer template throws error when querying elements because the ECSchemaRpcInterface is not registered. Also add bing maps integration.


aruniverse (approved):

ben-polinsky:
Tested locally FYI

PR #342, opened 2025-01-06
Viewer@5.0

aruniverse:
# iTwin Viewer 5.0

### Removed

The following prop(s) have been removed from the  `<Viewer />` component:

- `theme`
  - please wrap the `<Viewer />` component with `ThemeProvider` from `@itwin/itwinui-react` instead

- `backstageItems`
  - please provide a `Backstage` through [UiItemsProvider.provideBackstageItems](https://www.itwinjs.org/reference/appui-react/uiprovider/uiitemsprovider/).

- `getSchemaContext`
  - `schemaContext` has been added to `iModelConnection`

We have also dropped support for `CommonJs` modules. Please use `ESM` instead.

### Updated minimum versions

- The minimum `iTwin.js Core` versions have been updated to `^5.0.0`, please see their [changelog](https://github.com/iTwin/itwinjs-core/blob/master/docs/changehistory/5.0.0.md) for more details.
- The minimum `AppUI` versions have been updated to `^5.5.0`, please see their [changelog](https://github.com/iTwin/appui/releases) for more details.
- The minimum `imodels-*` versions have been updated to `^6.0.1`, please see their [changelog](https://github.com/iTwin/imodels-clients/blob/main/itwin-platform-access/imodels-access-frontend/CHANGELOG.md) for more details.
- The minimum `electron` version for the desktop-viewer has been updated to `^36.0.0`, please see their [changelog](https://www.electronjs.org/docs/latest/breaking-changes) for more details.
- The minimum version of `redux` and `react-redux` has been updated to `^5.0.0` and `^9.0.0` respectively. Please see their respective changelogs for more details.
- Now deliver `ES2023`


packages/apps/desktop-viewer-test/package.json
```diff
@@ -32,51 +32,54 @@
...
     "@itwin/itwinui-variables": "^2.0.0",
-    "@itwin/measure-tools-react": "^0.23.2",
-    "@itwin/presentation-backend": "^4.7.3",
-    "@itwin/presentation-common": "^4.7.3",
-    "@itwin/presentation-components": "^5.0.0",
-    "@itwin/presentation-frontend": "^4.7.3",
+    "@itwin/measure-tools-react": "latest",
+    "@itwin/object-storage-azure": "^3.0.0",
+    "@itwin/object-storage-core": "^3.0.0",
+    "@itwin/object-storage-google": "^3.0.0",
```
grigasp, line 71: I believe these 3 are not required with the latest release of `@itwin/imodels-access-backend`

grigasp: Also applies to `@itwin/cra-template-desktop-viewer` template

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+import React from "react";
+
+import { getAndSetViewState } from "../../../services/iModel/index.js";
+import type {
+  BlankConnectionViewState,
+  ViewerViewCreator3dOptions,
+  ViewerViewportControlOptions,
+} from "../../../types.js";
+
+const UnifiedSelectionViewport = viewWithUnifiedSelection(ViewportComponent); // eslint-disable-line @typescript-eslint/no-deprecated
```
grigasp, line 23: The previous implementation of `DefaultContentGroupProvider` was choosing a viewport component based on whether the Viewer was supplied a selection storage or not.
- When supplied, we were using `ViewportComponent`.
- When not supplied, we were using `viewWithUnifiedSelection(ViewportComponent)`, with deprecated `viewWithUnifiedSelection` providing unified selection functionality to the viewport.

Now we're always using `viewWithUnifiedSelection(ViewportComponent)` - is this change intentional?

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
```diff
@@ -39,19 +40,15 @@ const IModelLoader = React.memo((viewerProps: ModelLoaderProps) => {
...
   const { error, connection } = useConnection(viewerProps);
 
   const providers = useMemo<UiItemsProvider[]>(() => {
     const providers = [...(uiProviders || [])];
-    if (backstageItems?.length) {
-      providers.push(new BackstageItemsProvider(backstageItems));
-    }
     return providers;
-  }, [uiProviders, backstageItems]);
+  }, [uiProviders]);
```
ben-polinsky, line 51: Not sure if we care, but `uiProviders` being an array this useMemo will always re-evaluate, no?

ben-polinsky (approved):
Didn't review selection storage, but otherwise LGTM.

packages/modules/viewer-react/src/components/iModel/IModelLoader.tsx
ben-polinsky: Ah yes, thanks.

packages/modules/viewer-react/src/index.scss
saskliutas, line 1: Don't see that this file would be imported anywhere. Can we remove it?

saskliutas (approved):
Looks good

packages/modules/viewer-react/src/components/app-ui/providers/DefaultContentGroupProvider.tsx
aruniverse: Good catch. I switched it back to using just the `ViewportComponent`.
Personally, dont want to maintain both, and would prefer to drop deprecated behavior if possible. LMK your thoughts

aruniverse: I attempted to internal the slectionStorage and cleaned up the code. Can you review my last commit and if theres additional things that need to be cleaned up can you call out where or tackle it? I know theres some leftover backwards compat stuff in packages/modules/viewer-react/src/hooks/useUnifiedSelectionScopes.tsx

grigasp: > Do you see value in exposing selectionStorage as a prop?

Yes. An app that uses `<Viewer />` will most likely have components that also want to hook into unified selection, and we need all such components to use the same selection storage ([example](https://github.com/iTwin/viewer-components-react/blob/2548a657148b0b052b28b1c3b838ba76b3e201fb/apps/test-viewer/src/UiProvidersConfig.tsx#L126)).

> any use cases why someone wouldnt use the default `createStorage()` from the `unified-selection` package?

No, I don't see why anyone would need that. But as explained above, it's about having access to the storage.

I think we should have the prop as required, or have it optional and have fallbacks to the deprecated `Presentation.selection`.

PR #343, opened 2025-01-06
update to appui 5

aruniverse:
todo in another pr:
- [ ] revisit  deprecated appui apis exposed in our types
- [ ] fix tests (but we will need to move to vitest, so defer)
- [ ] review if we need our own ui providers or if should just appui's


packages/modules/viewer-react/src/components/app-ui/providers/ViewerNavigationToolsProvider.tsx
```diff
@@ -29,122 +29,12 @@ export class ViewerNavigationToolsProvider extends StandardNavigationToolsProvid
...
-          ToolbarHelper.createToolbarItemFromItemDef(
-            20,
-            CoreTools.toggleCameraViewCommand
-          )
-        );
-      }
-    }
-    return items;
+  // todo replace walk view tool with look and move tool
+  override getToolbarItems(): readonly ToolbarItem[] {
```
GerardasB, line 37: Missing functionality or the override can be removed entirely if no longer needed.

packages/modules/viewer-react/src/types.ts
```diff
@@ -61,7 +61,7 @@ export interface ViewerViewCreator3dOptions extends ViewCreator3dOptions {
 
 export interface ViewerFrontstage {
   /** frontstage provider to register */
-  provider: FrontstageProvider;
+  provider: FrontstageProvider; //eslint-disable-line deprecation/deprecation
```
GerardasB, line 64: Depending on your API versioning policy you might want to introduce: `frontstage: Frontstage` eventually as a replacement for deprecated `provider`.

GerardasB (approved):

ben-polinsky (approved):

packages/modules/viewer-react/src/components/app-ui/providers/ViewerNavigationToolsProvider.tsx
aruniverse: need to circle back to this. the viewer always used the LookAndMove tool as thats what Design Review uses. Wonder if the viewer should just use what AppUI uses or DR. further discussion needed

packages/modules/viewer-react/src/types.ts
aruniverse: yes, need to review this separately to see how we propagate AppUI deprecations to end users
Not sure if we want to provide backwards compat for now or just move to the new API. 
Further discussion needed and will revisit in separate PR as needed

Issue #344, opened 2025-01-09
Stop using selection-related APIs from `@itwin/presentation-frontend`


grigasp:
At the moment, the Viewer uses [SelectionScopesManager](https://github.com/iTwin/viewer/blob/66dd6b0014f1359baf36f2b51df097d32c2fc738/packages/modules/viewer-react/src/hooks/useUnifiedSelectionSync.tsx#L57) to get active selection scope. The API is [about to be deprecated](https://github.com/iTwin/itwinjs-core/issues/7527) - we should make sure consumers of the Viewer can properly specify the active scope without using deprecated APIs.

PR #346, opened 2025-02-21
Stop using deprecated selection APIs

grigasp:
Closes https://github.com/iTwin/viewer/issues/344. Also reduces usage of deprecated AppUI APIs.

For selection scopes we now have 3 options for consumers:

- **Option 1:** Don't supply `selectionScopes` prop.

    We default to using `Presentation.selection.scopes` for getting available and active scopes. 
    
    If AppUI's selection scopes' picker is used, it will be used to set the active scope both in the app's state, and on `Presentation.selection.scopes` for backwards compatibility.

    This option is basically just for backwards compatibility.

- **Option 2:** Supply `selectionScopes` prop with only `available` and `active` properties.

    The `active` prop is used to determine the initially active scope by looking up the `available` map.

    If AppUI's selection scopes' picker is used, it's populated with scopes from `available` map and the initially selected value is based on the `active` prop. When another value is selected, the default scope change handler changes the active scope in the app's state and on `Presentation.selection.scopes` for backwards compatibility.

    This option is for consumers who don't care about the active scope, as long as it's applied when selecting elements in graphics view and can be changed through selection scopes picker in the toolbar.

- **Option 3:** Supply `selectionScopes` prop with all properties, including `onChange`.

    In this case, it's consumer's responsibility to update the active scope by setting a new value.

    This option is for consumers who want to have full control over the active scope. The source of truth for it in on consumer's side, so consumer can use it in multiple components and change it whenever they want, using whatever component they want.


packages/modules/viewer-react/src/types.ts
```diff
@@ -113,7 +142,7 @@ export interface LoaderProps {
   loadingComponent?: React.ReactNode;
 }
 
-export type ViewerCommonProps = ViewerInitializerParams & LoaderProps & UnifiedSelectionProps;
+export type ViewerCommonProps = ViewerInitializerParams & LoaderProps & (UnifiedSelectionProps | { [Property in keyof UnifiedSelectionProps]?: never });
```
ben-polinsky, line 145: Perhaps extract this type since it's a bit unclear at first glance. Maybe an `AllOrNone<T>` utility type?

grigasp: 1. Extracted selection scope from `UnifiedSelectionProps` into a separate `SelectionScopeProps` type.
2. Replaced usages of `UnifiedSelectionProps | { [Property in keyof UnifiedSelectionProps]?: never }` with `AllOrNone<UnifiedSelectionProps>`.
3. Moved `UnifiedSelectionProps` from `ViewerCommonProps` and `ModelLoaderProps` into `LoaderProps`, which is base for both of them.

saskliutas (approved):

packages/modules/viewer-react/src/types.ts
grigasp: @ben-polinsky mind resolving the comment and approving the PR if you got no other suggestions?

aruniverse:
Do we expect all apps to make similar to changes to this? Wonder if we can extract some of this out to the unified-selction-react pkg, but I guess you are trying to avoid presentation-frontend dep.

cc @jjbeckman13 for web and @kckst8 @johnnyd710 for desktop apps

Does option 3 here provide a path forward for the issue described in https://github.com/iTwin/presentation/issues/767 ?

grigasp:
> Do we expect all apps to make similar to changes to this? Wonder if we can extract some of this out to the unified-selction-react pkg, but I guess you are trying to avoid presentation-frontend dep.

Most of the complexity is due to me trying to keep backwards compatibility - in multiple places you can see that I handle selection storage if it's provided and then fall back to `Presentation.selection` if not. That also involves mapping of new vs legacy types.

For apps it's much simpler:
1. Define a list of supported scopes like here: https://github.com/iTwin/viewer/blob/c02049469f7a6642d9706e3d0c65a4853afa12b4/packages/apps/web-viewer-test/src/components/home/ViewerHome.tsx#L262-L275
2. Supply the necessary props to selection scopes status bar item: https://github.com/iTwin/viewer/blob/c02049469f7a6642d9706e3d0c65a4853afa12b4/packages/modules/viewer-react/src/components/app-ui/providers/ViewerStatusbarItemsProvider.tsx#L175-L219
3. Supply the `activeScopeProvider` prop to `enableUnifiedSelectionSyncWithIModel`: https://github.com/iTwin/viewer/blob/c02049469f7a6642d9706e3d0c65a4853afa12b4/packages/modules/viewer-react/src/hooks/useUnifiedSelectionSync.tsx#L60

> Does option 3 here provide a path forward for the issue described in [iTwin/presentation#767](https://github.com/iTwin/presentation/issues/767) ?

Not really. As stated [in this comment](https://github.com/iTwin/presentation/issues/767#issuecomment-2464671388): "User workflows need to be investigated, considering both - read-only and the new read-write - scenarios".

grigasp:
@aruniverse, FYI I made some improvements to the migration guide in unified selection package, targeting specifically selection scopes: https://github.com/iTwin/presentation/blob/840e23a34cf9964e93929d869736ab3d53bcf5f7/packages/unified-selection/learning/MigrationGuide.md#appui-integration. This PR is used as a base for example code there, but without all the backwards-compatibility and configuration stuff.

aruniverse (approved):

aruniverse:
fyi @kckst8 @johnnyd710 - I imagine we'd want to make similar changes on your end

PR #357, opened 2025-05-03
Removing `getSchemaContext`

pankhur94:
https://github.com/iTwin/itwinjs-core/pull/7977 added `schemaContext` to `iModelConnection`, no longer require it as a prop in the viewer


packages/modules/viewer-react/src/hooks/useUnifiedSelectionSync.tsx
```diff
@@ -24,19 +23,17 @@ type SelectionScope = ReturnType<
...
   selectionStorage,
-  getSchemaContext,
 }: UseUnifiedSelectionSyncProps) {
   React.useEffect(() => {
-    if (!iModelConnection || !selectionStorage || !getSchemaContext) {
+    if (!iModelConnection || !selectionStorage) {
       return;
     }
-    const schemaContext = getSchemaContext(iModelConnection);
+    const schemaContext = iModelConnection.schemaContext;
```
aruniverse, line 36:
```suggestion
    const { schemaContext } = iModelConnection;
```

packages/apps/web-viewer-test/src/selectionStorage.ts
```diff
@@ -19,8 +18,7 @@ const imodelSchemaContextsCache = new Map<string, SchemaContext>();
 function getSchemaContext(imodel: IModelConnection) {
```
aruniverse, line 18: @saskliutas as soon as the presentation pkgs support 5, we should be able to drop this function
do you guys need the cache or is that just an optimization?

packages/modules/viewer-react/src/types.ts
```diff
@@ -83,10 +82,8 @@ export interface ViewerViewportControlOptions
 }
 
 export interface UnifiedSelectionProps {
-  /** Unified selection storage to synchronize with. Requires `getSchemaContext` prop to also be supplied. */
+  /** Unified selection storage to synchronize with. */
   selectionStorage?: SelectionStorage;
-  /** Function for getting schema context for an iModel. */
-  getSchemaContext?: (imodel: IModelConnection) => SchemaContext;
```
aruniverse, line 89: please note the removal of this in `iTwin/viewer/releases/CHANGELOG-5.0.md`

ben-polinsky (approved):

hl662 (approved):
I assume the deprecation errors will be tackled in a separate PR?

packages/apps/web-viewer-test/src/selectionStorage.ts
saskliutas: No, this cache was introduced only to make sure that all our component utilizing `SchemaContext` would use the same instance instead of each loading same schemas multiple times.

pankhur94: thanks. I have removed the functions.

packages/modules/viewer-react/src/types.ts
pankhur94: done

aruniverse (approved):

packages/apps/desktop-viewer-test/src/frontend/components/routes/ViewerRoute.tsx
```diff
@@ -69,7 +68,7 @@ export const ViewerRoute = () => {
                   getLabel: () => ModelsTreeComponent.getLabel(),
                   render: (props) => (
                     <ModelsTreeComponent
-                      getSchemaContext={getSchemaContext}
+                      getSchemaContext={(iModel) => iModel.schemaContext}
```
grigasp, line 71: Adding @saskliutas to confirm, but I think we're expecting this function to be memoized using `React.useCallback`, otherwise it will cause excessive re-renders.

saskliutas: It should be ok at this level as `render` is called only when tree is changed from selector.

packages/templates/cra-template-desktop-viewer/template/src/frontend/components/routes/ViewerRoute.tsx
```diff
@@ -115,7 +114,7 @@ export const ViewerRoute = () => {
       ]}
       enablePerformanceMonitors={true}
       selectionStorage={unifiedSelectionStorage}
-      getSchemaContext={getSchemaContext}
+      getSchemaContext={(iModel) => iModel.schemaContext}
```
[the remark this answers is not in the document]
pankhur94: yes. that is right

packages/templates/cra-template-web-viewer/template/src/App.tsx
```diff
@@ -234,7 +234,7 @@ const App: React.FC = () => {
           new MeasureToolsUiItemsProvider(),
         ]}
         selectionStorage={unifiedSelectionStorage}
-        getSchemaContext={getSchemaContext}
+        getSchemaContext={(iModel) => iModel.schemaContext}
```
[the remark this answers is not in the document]
pankhur94: yes that is right.

aruniverse:
We needed to also register the ECSchemaRpcInterface by default with this PR, I forgot. I have added it in this commit, https://github.com/iTwin/viewer/pull/342/commits/959a5d206e4c3aa98e63e0c76e70e1feeee77785

PR #358, opened 2025-05-06
Fixing lint for CI pass

pankhur94:
This PR addresses and resolves existing linting errors across the codebase. In addition, it updates several ESLint-related dependencies to align with our current setup and improve compatibility.
Changes Include
- Fixed lint errors reported by @typescript-eslint/parser
- Updated ESLint configuration files to move to flat file format.
- Upgraded the lint-related dependencies.


packages/apps/web-viewer-test/.env
```diff
@@ -19,7 +19,7 @@ IMJS_AUTH_CLIENT_ITWIN_ID_PROD="1bff8c44-3196-4231-b8f6-66cf6dacd45b"
 IMJS_AUTH_CLIENT_IMODEL_ID_PROD="71adc398-33bd-4ca9-9dec-fa6a74729bf6"
 IMJS_AUTH_CLIENT_IMODEL_ID2_PROD="435e704b-ae19-418c-a02e-fc75686e213c"
 
-IMJS_AUTH_CLIENT_CLIENT_ID=""
+IMJS_AUTH_CLIENT_CLIENT_ID="spa-QtXdyYDarj96WoCNBj8fQCA7V"
```
aruniverse, line 22: dont commit this to src control
your overrides should go into a `.env.local` file

packages/modules/desktop-viewer-react/src/services/Initializer.ts
```diff
@@ -56,7 +56,6 @@ export class DesktopInitializer {
         iModelAppOpts.authorizationClient = authClient;
         ViewerAuthorization.client = authClient;
 
-        // eslint-disable-next-line deprecation/deprecation
```
aruniverse, line 59: not deprecated anymore?

packages/modules/desktop-viewer-react/tsconfig.json
```diff
@@ -1,10 +1,8 @@
 {
-  "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
+  "extends": "../../../common/config/tsconfig/tsconfig.base.json",
```
aruniverse, line 2: any reason to switch this?
would prefer we use the one from the build-tools pkg as a default for everything in `modules/` 
not a big deal to change in the `apps/`

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
```diff
@@ -90,12 +90,14 @@ export class ViewCreator3d extends ViewCreator {
               }, 100);
             });
           };
-
-          tileTreesLoaded().finally(() => {
+           
+          tileTreesLoaded().then(() => {
```
aruniverse, line 94: we want this behavior to run even when the promise rejects (due to timeout)

common/scripts/eslint.base.config.js
```diff
@@ -0,0 +1,54 @@
+const tsPlugin = require("@typescript-eslint/eslint-plugin");
+const tsParser = require("@typescript-eslint/parser");
+const react = require("eslint-plugin-react");
+const prettier = require("eslint-plugin-prettier");
+const simpleImportSortPlugin = require("eslint-plugin-simple-import-sort");
+const reactHooks = require("eslint-plugin-react-hooks");
+
+module.exports = [
```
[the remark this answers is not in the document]
pankhur94: good point, I have changed it to a singular object in all the configs.

common/scripts/eslint.config.js
```diff
@@ -0,0 +1,14 @@
+const baseConfig = require("./eslint.base.config.js");
+
+module.exports = [
```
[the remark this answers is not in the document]
pankhur94: done

common/scripts/package.json
```diff
@@ -5,23 +5,24 @@
...
-    "@typescript-eslint/parser": "^4.5.0",
-    "eslint": "^7.11.0",
+    "@typescript-eslint/eslint-plugin": "^8.32.0",
+    "@typescript-eslint/parser": "^8.32.0",
+    "eslint": "~9.10.0",
     "eslint-config-airbnb": "^18.2.1",
     "eslint-config-prettier": "^6.11.0",
     "eslint-config-react-app": "^6.0.0",
-    "eslint-plugin-flowtype": "^5.2.0",
+    "eslint-plugin-flowtype": "^8.0.3",
```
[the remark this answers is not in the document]
pankhur94: in order for the plugins to be compatible with eslint v9, we had to update eslint-plugin and parser, as the older version did not follow flat file rules and used extends, which is not supported in v9 eslint.
I have removed flowtype and added itwin/eslint-plugin.

packages/apps/web-viewer-test/.env
pankhur94: this is removed, and I have added overrides in .env.local which will not be commited.

packages/modules/desktop-viewer-react/src/services/Initializer.ts
pankhur94: I  have added no-deprecate rule now. and have suppressed it wherever necessary. We will probably take up updating deprecation in a separate pr.

packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts
pankhur94: reverted to using finally. Have added a comment to supress the warning.

packages/modules/desktop-viewer-react/tsconfig.json
pankhur94: I noticed that when extending from our build-tools, our compiler options' target and lib fields are not overriden. However, it does when I extend it from a local common file(exactly similar to build-tools' base config) it works. Not sure why this happens.

common/scripts/package.json
```diff
@@ -5,23 +5,24 @@
   "license": "MIT",
   "repository": "https://github.com/itwin/viewer/tree/main/common/scripts",
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^4.5.0",
-    "@typescript-eslint/parser": "^4.5.0",
-    "eslint": "^7.11.0",
+    "@typescript-eslint/eslint-plugin": "^8.32.0",
+    "@typescript-eslint/parser": "^8.32.0",
+    "@itwin/eslint-plugin": "^5.1.0",
+    "eslint": "~9.10.0",
```
aruniverse, line 11: any reason youre pinning it down to this minor?

common/scripts/eslint.config.js
```diff
@@ -0,0 +1,12 @@
+const baseConfig = require("./eslint.base.config.js");
+
+module.exports = {
+  ...baseConfig,
+  rules: {
+    ...(baseConfig.rules || {}),
```
aruniverse, line 6: any reason for spreading the empty object? 
wouldnt baseConfig.rules always be defined?

common/scripts/package.json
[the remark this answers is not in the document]
aruniverse: none of the other `eslint-plugin-*` pkgs didnt need to get updated?

packages/modules/viewer-react/tsconfig.json
```diff
@@ -1,11 +1,9 @@
 {
-  "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
+  "extends": "../../../common/config/tsconfig/tsconfig.base.json",
   "compilerOptions": {
-    "allowSyntheticDefaultImports": true,
-    "esModuleInterop": true,
-    "target": "ES2023",
-    "lib": ["ES2023", "DOM", "DOM.Iterable", "ESNext.Disposable"]
+    "target": "ES2022",
```
aruniverse, line 4: any reason you are now downgrading the target from `es2023` to `es2022`?

packages/modules/viewer-react/tsconfig.test.json
```diff
@@ -0,0 +1,8 @@
+{
```
aruniverse, line 1: as mentioned in another comment, try to stay consistent with the rest of the repo you are working in
everything else is using 2 spaces for indents, you should as well

common/scripts/package.json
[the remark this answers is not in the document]
aruniverse: many of these deps here seem outdated
can verify with `rush-pnpm outdated`

packages/modules/desktop-viewer-react/tsconfig.json
aruniverse: any reason to override the target?
i would prefer we investgate why it doesnt work as we expect vs creating a new config

when we publish these pkgs now, youre gonna have to dig through our repo to find this config since we dont publish this config; whereas previously it was self contained to some extent

common/scripts/package.json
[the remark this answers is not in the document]
pankhur94: I have updated all the deps. Some of the deps have been removed as they weren't being used.

common/scripts/package.json
pankhur94: I noticed that we use ^9.13.0 in itwinjs-core and I have updated it.

packages/modules/viewer-react/tsconfig.test.json
pankhur94: done

common/scripts/eslint.config.js
pankhur94: you are right. updated.

packages/modules/desktop-viewer-react/tsconfig.json
pankhur94: The core issue appears to be with the target and lib settings in our tsconfig.json. For some reason, es2023 is not being accepted as a valid value, and we see the following error:
image
This typically points to an older version of TypeScript being used. However, we're currently using @typescript-eslint/parser@^8.52, which has a dependency on a newer TypeScript version that should support es2023.

It seems like there's a mismatch somewhere — possibly ESLint or one of its plugins is resolving a different (older) TypeScript version. This has us stuck for now.

packages/modules/viewer-react/tsconfig.json
pankhur94: see this(https://github.com/iTwin/viewer/pull/358#discussion_r2080867032)

packages/modules/desktop-viewer-react/tsconfig.json
pankhur94: I found out that in `desktop-viewer-test` package, we were using a different minor version of ts i.e. ~5.0 and this was causing the issue with not allowing es2023. Although I am not sure why eslint parser picks this version of ts. I still see there are couple of more packages which use older version of ts(they dont effect the parser for some reason) like `error-handling-react` and `web-viewer-test`. Were these not updated intentionally? If not I can update them as well.

aruniverse: we don't use `error-handling-react` anymore (take a look at rush.json to see what pkgs we currently maintain), its just there for archival purposes.

no, its probably mistake, I would recommend updating things and making sure they are consistent across the repo and that things work

packages/modules/viewer-react/tsconfig.test.json
```diff
@@ -0,0 +1,8 @@
...
+  "extends": "./tsconfig.json",
```
aruniverse, line 2: any reason we created this new test tsconfig?

aruniverse: I had explicitly excluded tests temporarily to not have the fix all the tests while there were breaking changes going on. We should be getting to a point where we can remove the exclusion and update the tests as needed (in a separate pr ofc). So I think you can remove this config

pankhur94: tsconfig.test has been deleted. I have excluded tests in eslint(to avoid errors) for now. Once we remove exclusion of tests from tsconfig.json, we can remove it from eslint.config.

PR #359, opened 2025-05-13
[Fix]: iTwin Viewer template issue

pankhur94:
- Added "@itwin/unified-selection-react" deps

### DIFF
--- a/common/changes/@itwin/cra-template-desktop-viewer/pankhur94-adding-unified-react-deps_2025-05-13-17-20.json
+++ b/common/changes/@itwin/cra-template-desktop-viewer/pankhur94-adding-unified-react-deps_2025-05-13-17-20.json
@@ -2,8 +2,8 @@
   "changes": [
     {
       "packageName": "@itwin/cra-template-desktop-viewer",
-      "comment": "",
-      "type": "none"
+      "comment": "Adding unified-react dependency to desktop viewer template.",
+      "type": "patch"
     }
   ],
   "packageName": "@itwin/cra-template-desktop-viewer"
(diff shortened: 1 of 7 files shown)


grigasp:
Why is it required?

pankhur94:
> Why is it required?

it was needed because it is noted as a peer dependency in `@itwin/presentation-components` and it was leading to the issue that I have linked above.

common/changes/@itwin/cra-template-desktop-viewer/pankhur94-adding-unified-react-deps_2025-05-13-17-20.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/cra-template-desktop-viewer",
+      "comment": "",
```
ben-polinsky, line 5: Please add comment and "type": "patch" to both change entries.

aruniverse (approved):

grigasp:
> > Why is it required?
> 
> `@itwin/unified-selection-react` was needed because it is noted as a peer dependency in `@itwin/presentation-components` and it was leading to the issue that I have linked above.

The peer is marked `optional`:
```json
  "peerDependenciesMeta": {
    "@itwin/unified-selection-react": {
      "optional": true
    }
  },
```

And the code that emits the warning is aware of `@itwin/unified-selection-react` possibly not being available: https://github.com/iTwin/presentation/blob/af6ab3d446b501662c0a03f4b94721ddbc4ea631/packages/components/src/presentation-components/common/UnifiedSelection.tsx#L29-L37. We should look into how to get rid of the warning instead.

ben-polinsky:
> We should look into how to get rid of the warning instead. We're using webpack, right?

Agree in theory, but not sure it's worth it at this point as we look toward a starter that isn't an app template but rather a starter repo (backed by vite).

PR #361, opened 2025-05-16
[Fix] : Template issue

pankhur94:
Removes `@itwin/presentation-core-interop` as direct dep and override for metadependencies to 1.3.1.

### DIFF
--- a/common/changes/@itwin/cra-template-web-viewer/pankhur94-template-bug-fix_2025-05-16-22-06.json
+++ b/common/changes/@itwin/cra-template-web-viewer/pankhur94-template-bug-fix_2025-05-16-22-06.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/cra-template-web-viewer",
+      "comment": "remove direct dep @itwin/presentation-core-interop and add override for metadependencies to 1.3.1",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/cra-template-web-viewer"
+}
--- a/packages/templates/cra-template-web-viewer/template.json
+++ b/packages/templates/cra-template-web-viewer/template.json
@@ -27,7 +27,6 @@
       "@itwin/presentation-common": "^4.7.3",
       "@itwin/presentation-components": "^5.0.0",
       "@itwin/presentation-frontend": "^4.7.3",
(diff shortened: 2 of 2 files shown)


pankhur94:
I could not find any instance of direct usage of the package `@itwin/presentation-core-interop` directly in the app and hence took the liberty to remove it. If we need it as a direct dep, we would have to ask the user to manually change it to 1.3.1 as bentley react scripts prefixes "^" which leads to conflict with the override. The user would then have to manually add below in package.json as well before npm install.

```json
overrides : {
"@itwin/presentation-core-interop": "1.3.1"
}
```

aruniverse:
`@itwin/presentation-core-interop` must be a peer dep of some other pkg, which is why we need to include it in our end
Review the blame here: https://github.com/iTwin/viewer/blame/master/packages/templates/cra-template-web-viewer/template.json#L30

pankhur94:
> `@itwin/presentation-core-interop` must be a peer dep of some other pkg, which is why we need to include it in our end Review the blame here: https://github.com/iTwin/viewer/blame/master/packages/templates/cra-template-web-viewer/template.json#L30

I did check and found out that it was added in this [commit](https://github.com/iTwin/viewer/commit/d7e7ad18dfb36bacb0f16157261ff11c2f6e7109#diff-7efe8b82b9f39cf3f51f72072af1563d6703da6d20040bab73df33f1d50ee34f). However I could not find any of the other dependencies added/updated to have it as a peer-dependency. To confirm I ran npm ls @itwin/presentation-core-interop --all in the newly created app and did not find `@itwin/presentation-core-interop` to be a peer dependency in any of the deps either.
image
Is there any better/refined way to search for peer-dependencies?

grigasp:
I agree, it looks like it's unnecessary, I think adding it was just an oversight.

However, I don't understand why the following part is necessary:
```json
    "overrides": {
      "@itwin/presentation-core-interop": "1.3.1"
    }
```
Why don't we want to let consumers choose which version to use?

ben-polinsky (approved):

aruniverse:
But how are these overrides applied? does npm automatically apply them?

pankhur94:
> But how are these overrides applied? does npm automatically apply them?

yes, that is right, npm understands overrides and when it does npm install, it applies them.

PR #366, opened 2025-05-29
Attempt to clean up dependencies

grigasp:
Cleaned up dependencies in attempt to resolve peer deps mess.

Identified issues:

1. `@itwin/electron-authorization` and `@itwin/browser-authorization` expose a class (`BeEvent`) from `@itwin/core-bentley`, which is a direct dependency, through their public APIs. This should be a peer dep instead. For now I fixed this through `.pnpmfile.cjs`, but it needs to be fixed in those packages.

2. The latest `@itwin/electron-authorization` has a peer dep on `electron @ >=23.0.0 <34.0.0`, but `itwinjs-core @ 5.0-rc` expects at least `35.0`.


aruniverse:
cc https://github.com/iTwin/auth-clients/pull/301#issuecomment-2920581862

PR #368, opened 2025-06-12
[ Fix ] : Desktop and Web Viewer Test App

pankhur94:
- Blank Connection is always closed and this leads to errors when getting scopes. Added an isOpen check.
- instanceof is not behaving reliably, replaced it with a custom type check ElectronRendererAuthorization.


pankhur94:
not putting this up for review as have not been able to test desktop viewer app since itwinjs-core changes is breaking viewer.

packages/apps/desktop-viewer-test/src/frontend/components/signin/SignIn.tsx
```diff
@@ -6,17 +6,18 @@
...
+import { isElectronRendererAuth } from "../../util/typeCheck";
+
 export const SignIn = () => {
   const [signingIn, setSigningIn] = useState(false);
 
   const onSignInClick = async () => {
     setSigningIn(true);
     if (
-      IModelApp.authorizationClient instanceof ElectronRendererAuthorization
+      isElectronRendererAuth(IModelApp.authorizationClient)
```
ben-polinsky, line 20: Why was the `instanceof` check not working?

packages/modules/viewer-react/src/hooks/useUnifiedSelectionScopes.tsx
```diff
@@ -31,7 +31,7 @@ export function useUnifiedSelectionScopes({
     }
 
     let isDisposed = false;
-    if (iModelConnection) {
+    if (iModelConnection && iModelConnection.isOpen) {
```
ben-polinsky, line 34:
```suggestion
    if (iModelConnection?.isOpen) {
```

packages/apps/desktop-viewer-test/src/frontend/components/signin/SignIn.tsx
pankhur94: instanceof only returns true if the object was created by directly instantiating the class (or a subclass of it).
So, for IModelApp.authorizationClient instanceof ElectronRendererAuthorization to return true, the value must have been assigned as:
```
IModelApp.authorizationClient = new ElectronRendererAuthorization();
```
 If ElectronRendererAuthorization.prototype exists in the chain of authorizationClient, the expression evaluates to true.

ben-polinsky: This does happen eventually. We create the instance, it passes through a bunch of options, and eventually lands in iModelApp.ts in core and is set. It should be in the prototype chain?

ben-polinsky: Where do we go over IPC?

ben-polinsky: I'm not able to run this branch because of conflicting presentation modules

ben-polinsky: Yeah I can replicate, but I am still puzzled... I'll resolve for sake of time.

ben-polinsky (approved):

saskliutas:
> > not putting this up for review as have not been able to test desktop viewer app since itwinjs-core changes is breaking viewer.
> 
> Tested desktop app by pinning to version `@itwin/presentation-frontend@5.0.0-dev.117` and it seems to be working as expected. Included the necessary template changes as well.

Took a quick look at why `presentation-frontend` causes problems in desktop-viewer-test. It looks like desktop-viewer-test tries to load both `esm` and `cjs` versions of `presentation-frontend`. This worked before because there were no checks to make sure only one `presentation-frontend` version is loaded. 

This is a valid issue and it should be fixed in desktop-viewer-test instead of pinning older `presentation-frontend` version.

pankhur94:
> > > not putting this up for review as have not been able to test desktop viewer app since itwinjs-core changes is breaking viewer.
> > 
> > 
> > Tested desktop app by pinning to version `@itwin/presentation-frontend@5.0.0-dev.117` and it seems to be working as expected. Included the necessary template changes as well.
> 
> Took a quick look at why `presentation-frontend` causes problems in desktop-viewer-test. It looks like desktop-viewer-test tries to load both `esm` and `cjs` versions of `presentation-frontend`. This worked before because there were no checks to make sure only one `presentation-frontend` version is loaded.
> 
> This is a valid issue and it should be fixed in desktop-viewer-test instead of pinning older `presentation-frontend` version.



> > > not putting this up for review as have not been able to test desktop viewer app since itwinjs-core changes is breaking viewer.
> > 
> > 
> > Tested desktop app by pinning to version `@itwin/presentation-frontend@5.0.0-dev.117` and it seems to be working as expected. Included the necessary template changes as well.
> 
> Took a quick look at why `presentation-frontend` causes problems in desktop-viewer-test. It looks like desktop-viewer-test tries to load both `esm` and `cjs` versions of `presentation-frontend`. This worked before because there were no checks to make sure only one `presentation-frontend` version is loaded.
> 
> This is a valid issue and it should be fixed in desktop-viewer-test instead of pinning older `presentation-frontend` version.

Upon investigation, I found that the issue stems from a [change](https://github.com/iTwin/viewer/pull/365/files#diff-717d5906f37d5f021d0dc38e1af68cef0a1864b05e70da2575291a72e8aac0d5:~:text=%22esModuleInterop%22%3A%20true%2C,moduleResolution%22%3A%20%22NodeNext%22%2C) introduced during the merge from master to next. This change was made because --module commonjs and --moduleResolution node do not respect the exports field in [`@itwin/electron-authorization`](https://github.com/iTwin/auth-clients/blob/main/packages/electron/package.json#L5-L16). Since our package is still marked as "type": "commonjs", using moduleResolution: nodenext causes TypeScript to treat all .ts files as CommonJS and emit CJS output — even when building for ESM. As a result, require calls are being generated for `@itwin/presentation-frontend`, which pulls in its CommonJS build and leads to the duplicate import issue.
To correctly generate ESM output while still respecting the exports field of dependencies, we can use a combination of module: "ES2020" and moduleResolution: "bundler". I tested this setup locally, and it works as expected. The reason I'm confident in using bundler is based on this note in the TypeScript docs:https://www.typescriptlang.org/tsconfig/#moduleResolution:~:text=%27bundler%27%20for%20use%20with%20bundlers.%20Like%20node16%20and%20nodenext%2C%20this%20mode%20supports%20package.json%20%22imports%22%20and%20%22exports%22%2C%20but%20unlike%20the%20Node.js%20resolution%20modes%2C%20bundler%20never%20requires%20file%20extensions%20on%20relative%20paths%20in%20imports.

Let me know if you have any thoughts or suggestions.

PR #370, opened 2025-06-16
[ Fix ] : Web & Desktop Template build issue

pankhur94:
- Version [8.15.0](https://github.com/acornjs/acorn/blob/master/acorn/CHANGELOG.md#8150-2025-06-08) of Acorn misinterprets a simple function imported from `@itwin/core-bentley` as a using await declaration, causing the build to fail.
- Pinning to version 8.14.0 until the [issue](https://github.com/acornjs/acorn/issues/1373) is fixed.


pankhur94:
> This is caused by a recent addition to `acorn` library to support `using` syntax: https://github.com/acornjs/acorn/blob/master/acorn/CHANGELOG.md#8150-2025-06-08. As far as I can tell, there's already an issue related to this: [acornjs/acorn#1373](https://github.com/acornjs/acorn/issues/1373).
> 
> Until that's fixed, a temporary workaround on our end could be adding an override to `package.json`:
> 
> ```json
> {
>   "overrides": {
>     "acorn": "~8.14.0"
>   }
> }
> ```

Interesting — I pinned acorn to 8.14.0 in both templates and things started working.

aruniverse (approved):

PR #371, opened 2025-06-17
[ Fix ] : Desktop viewer Test duplicate Presentation frontend

pankhur94:


saskliutas:
> if someone tries to use `@itwin/desktop-viewer-react` as `cjs` they will run into runtime issues because `require(ESM)`

Looks like same is reproducible by unit tests in this repo as they run as `cjs`. Running `@itwin/viewer-react` unit test produces errors and suggest enabling `ESM` modules in jest.

pankhur94:
> > if someone tries to use `@itwin/desktop-viewer-react` as `cjs` they will run into runtime issues because `require(ESM)`
> 
> Looks like same is reproducible by unit tests in this repo as they run as `cjs`. Running `@itwin/viewer-react` unit test produces errors and suggest enabling `ESM` modules in jest.

took a jab at dropping cjs [here](https://github.com/iTwin/viewer/pull/374). Trying to fix the tests.

PR #374, opened 2025-06-19
Drop cjs

pankhur94:
- Drop cjs in favour of esm for `@itwin/viewer-react, @itwin/web-viewer-react, @itwin/desktop-viewer-react`
- Convert jest to vitest for `@itwin/viewer-react, @itwin/web-viewer-react`
- Adding tests to run in ci

### DIFF
--- a/packages/modules/viewer-react/tsconfig.json
+++ b/packages/modules/viewer-react/tsconfig.json
@@ -7,5 +7,4 @@
     "moduleResolution": "NodeNext"
   },
   "include": ["./src/**/*.ts*"],
-  "exclude": ["./src/tests/**/*.ts*"]
 }
--- a/packages/modules/web-viewer-react/tsconfig.json
+++ b/packages/modules/web-viewer-react/tsconfig.json
@@ -7,5 +7,4 @@
     "moduleResolution": "NodeNext"
   },
   "include": ["./src/**/*.ts*"],
-  "exclude": ["./src/tests/**/*.ts*"]
 }
--- a/packages/modules/desktop-viewer-react/tsconfig.json
+++ b/packages/modules/desktop-viewer-react/tsconfig.json
@@ -7,5 +7,4 @@
     "moduleResolution": "NodeNext"
   },
   "include": ["./src/**/*.ts*"],
-  "exclude": ["./src/tests/**/*.ts*"]
 }
--- a/packages/modules/viewer-react/src/tests/components/BaseViewer.test.tsx
+++ b/packages/modules/viewer-react/src/tests/components/BaseViewer.test.tsx
@@ -7,8 +7,8 @@ import { render, waitFor } from '@testing-library/react';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import React from 'react';
 import { UiCore } from '@itwin/core-react';
-import { BaseViewer } from "../..";
-import * as IModelService from "../../services/iModel/IModelService";
+import { BaseViewer } from "../../components/BaseViewer.js";
+import * as IModelService from "../../services/iModel/IModelService.js";
 
 vi.mock("@itwin/presentation-frontend", async (importActual) => {
   const original = await importActual<typeof import("@itwin/presentation-frontend")>();
--- a/packages/modules/viewer-react/src/tests/services/BaseInitializer.test.ts
+++ b/packages/modules/viewer-react/src/tests/services/BaseInitializer.test.ts
@@ -12,8 +12,8 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 import {
   BaseInitializer,
   getIModelAppOptions,
-} from "../../services/BaseInitializer";
-import { MockToolAdmin } from "../mocks/MockToolAdmin";
+} from "../../services/BaseInitializer.js";
+import { MockToolAdmin } from "../mocks/MockToolAdmin.js";
 
 vi.mock("../../services/iModel/ViewCreator3d", () => {
   return {


pankhur94:
jest does not support testing esm fully and is still in experimental phase. Switching to vitest(fully esm compatible) for better esm testing.

packages/modules/desktop-viewer-react/package.json
```diff
@@ -2,9 +2,15 @@
   "name": "@itwin/desktop-viewer-react",
   "version": "5.0.0-dev.7",
   "description": "Configurable iTwin.js viewer for desktop applications",
-  "main": "lib/cjs/index.js",
+  "type": "module",
   "module": "lib/esm/index.js",
```
saskliutas, line 6: I think `"module": "lib/esm/index.js"` can be removed. Also I don't think we need `lib/esm` nesting and build output can be directly put under `lib` forlder.

packages/modules/desktop-viewer-react/package.json
```diff
@@ -13,8 +19,7 @@
   "repository": "https://github.com/itwin/viewer/tree/main/packages/modules/desktop-viewer-react",
   "scripts": {
     "start": "concurrently \"npm:build:cjs -- -w\" \"npm:build:esm -- -w\"",
-    "build": "npm run -s build:cjs && npm run -s build:esm",
-    "build:cjs": "tsc 1>&2 --outDir lib/cjs",
+    "build": "npm run -s build:esm",
     "build:esm": "tsc 1>&2 --outDir lib/esm",
```
saskliutas, line 23: Don't think there is a need for `build` and `build:esm` scirpts as they both do the same. Also `start` script still tries to run `build:cjs`.

packages/modules/viewer-react/package.json
```diff
@@ -2,9 +2,14 @@
   "name": "@itwin/viewer-react",
   "version": "5.0.0-dev.7",
   "description": "Configurable iTwin.js viewer",
-  "main": "lib/cjs/index.js",
+  "type": "module",
   "module": "lib/esm/index.js",
```
saskliutas, line 6: Same as in other package. I think `"module": "lib/esm/index.js"` can be removed and `lib/esm` can be changed to `lib`.

packages/modules/viewer-react/tsconfig.json
```diff
@@ -3,6 +3,9 @@
   "compilerOptions": {
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
+    "target": "es2020"
```
saskliutas, line 8: Why is `target` used only in this package tsconfig?

packages/modules/web-viewer-react/package.json
```diff
@@ -2,22 +2,28 @@
   "name": "@itwin/web-viewer-react",
   "version": "5.0.0-dev.7",
   "description": "Configurable iTwin.js viewer for web applications",
-  "main": "lib/cjs/index.js",
+  "type": "module",
   "module": "lib/esm/index.js",
```
saskliutas, line 6: Same as in other package. `"module": "lib/esm/index.js"` can be removed and `lib/esm` changed to `lib`.

packages/modules/viewer-react/package.json
```diff
@@ -13,18 +18,16 @@
...
-    "build-scss": "concurrently npm:dev-build-scss npm:prod-build-scss",
-    "dev-build-scss": "copyfiles -u1 \"./src/**/*.scss\" ./lib/cjs",
+    "build": "npm run build:esm && npm run build-scss && npm run copy-public",
+    "build:esm": "tsc 1>&2 --outDir lib/esm",
+    "build-scss": "npm run prod-build-scss",
     "prod-build-scss": "copyfiles -u1 \"./src/**/*.scss\" ./lib/esm",
     "copy-public": "copyfiles -u1 \"./public/**/*\" ./lib/public",
-    "test": "jest",
-    "test-watch": "jest --watch",
+    "vitest:test": "cross-env NODE_OPTIONS=\"--experimental-vm-modules \" vitest",
```
saskliutas, line 28: Why is `--experimental-vm-modules` flag is needed when running `vitest`?

saskliutas (changes requested):

packages/modules/desktop-viewer-react/package.json
pankhur94: done

packages/modules/desktop-viewer-react/package.json
pankhur94: fixed

packages/modules/desktop-viewer-react/package.json
```diff
@@ -2,20 +2,22 @@
   "name": "@itwin/desktop-viewer-react",
   "version": "5.0.0-dev.7",
   "description": "Configurable iTwin.js viewer for desktop applications",
-  "main": "lib/cjs/index.js",
-  "module": "lib/esm/index.js",
-  "types": "lib/cjs/index.d.ts",
```
saskliutas, line 7: You should leave `types` field to make sure this package works in apps with outdated tsconfig. Also `types` field adds `TS` badge in npm registry.

saskliutas:
> I noticed that `@itwin/desktop-viewer-react` includes a Jest configuration and an active test-watch command, but there are currently no tests defined for this package. Would it make sense to remove Jest and any related dependencies from it to reduce maintenance overhead.

I think `jest` and related dependencies should be removed. I guess there can be a case to add vitest config if there are plans to have unit tests in the future but I think this part can be handled when said test are being added.

packages/modules/desktop-viewer-react/package.json
pankhur94: added.

aruniverse:
@hl662 can you review the vitest changes?

packages/modules/viewer-react/src/tests/components/iModel/IModelLoader.test.tsx
```diff
@@ -270,41 +312,38 @@ describe("IModelLoader", () => {
...
     await waitFor(() => getByTestId("viewer"));
-
     expect(IModelServices.openLocalIModel).toHaveBeenCalledWith(
       "x://iModel",
       undefined
     );
   });
 
-  it("sets the theme to the provided theme", async () => {
+  it.skip("sets the theme to the provided theme", async () => {
```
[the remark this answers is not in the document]
pankhur94: removed

packages/modules/desktop-viewer-react/package.json
```diff
@@ -2,23 +2,26 @@
...
-    "start": "concurrently \"npm:build:cjs -- -w\" \"npm:build:esm -- -w\"",
-    "build": "npm run -s build:cjs && npm run -s build:esm",
-    "build:cjs": "tsc 1>&2 --outDir lib/cjs",
-    "build:esm": "tsc 1>&2 --outDir lib/esm",
+    "start": "npm run build -- -w",
+    "build": "tsc 1>&2 --outDir lib",
     "prod-build": "tsc -b",
     "test": "",
-    "test-watch": "jest --watch",
+    "test-watch": "",
```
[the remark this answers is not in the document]
pankhur94: removed both "test" and "test-watch". Can be added in future.

packages/modules/viewer-react/vitest.config.mts
```diff
@@ -0,0 +1,22 @@
...
+    globals: true,
+    include: ["src/tests/**/*.test.{ts,tsx}"],
+    server: {
+      deps: {
+        inline: ["@itwin/appui-react", "@itwin/imodel-components-react", "@itwin/components-react", "@itwin/core-react",
+          "@itwin/presentation-components"
+        ]
+      }
+    },
+    minWorkers: 1,
```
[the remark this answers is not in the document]
pankhur94: I see we use the same in itwinjs-core. How do we actually know what suits our needs?

hl662:
@aruniverse Should we update minSupportedNodeVersion in rush.json? Drop 18, and either match the version in itwin core or go further and do 20.19 || 22.12

packages/modules/viewer-react/vitest.config.mts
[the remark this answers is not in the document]
hl662: By default, maxWorkers uses all available cores in a CI pipeline, and with the number of tests in core (and on top of vitest, rush also executes each folder in parallel), we run into resource starvation. Seeing how this repo also uses rush I'd keep the min/maxWorkers like here too then

packages/modules/web-viewer-react/tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
   "compilerOptions": {
     "allowSyntheticDefaultImports": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext"
   },
   "include": ["./src/**/*.ts*"],
```
hl662, line 9: can you try to exclude all test files? Vitest shouldn't need a lib output to run the tests (compared to jest). We can reduce the file size in `lib` then

aruniverse:
> @aruniverse Should we update minSupportedNodeVersion in rush.json? Drop 18, and either match the version in itwin core or go further and do 20.19 || 22.12

we can do so in another pr. lets go with the latter in that

packages/modules/web-viewer-react/tsconfig.json
aruniverse: reducing file size in lib will make no difference, we dont ship the entirety of lib. see `.npmignore`.
the only benefit here would be saving a few seconds during compilation.

PR #379, opened 2025-07-10
Moving web-viewer-template from cra to vite

pankhur94:
- The issue we're currently facing is that adding dependencies directly to package.json causes rush install to fail, because we specify "@itwin/web-viewer-react": "^5.0.0" — which also exists within the Rush workspace.

- To work around this, I’ve added a custom script that copies dependencies from template.json into package.json before running npm install.

- I tested this using degit, and it works fine since degit simply clones the template without triggering rush or linking workspace packages.

- A cleaner and more automated approach would be to use a preinstall script that merges template.json into package.json, but only when the template is used as a standalone app (i.e., outside a monorepo). This avoids interfering with Rush workflows during local development.

To-Do:

- [ ] Update Readme.md


packages/templates/vite-template-web-viewer/vite.config.ts
```diff
@@ -0,0 +1,62 @@
...
+        {
+          // Resolve SASS tilde imports.
+          find: /^~(.*)$/,
+          replacement: "$1",
+        },
+      ],
+    },
+    envPrefix: ENV_PREFIX,
+    define: {
+      "process.env.IMJS_URL_PREFIX": env.IMJS_URL_PREFIX
```
aruniverse, line 57: we should prob move away from using process.env in these new templates...

aruniverse:
> > The issue we're currently facing is that adding dependencies directly to package.json causes rush install to fail, because we specify "@itwin/web-viewer-react": "^5.0.0" — which also exists within the Rush workspace.
> 
> Could you elaborate a bit? I don't think the extra scripting should be necessary.

I agree with ben re the scripting. But also, this makes no sense. We already re-use packages within the monorepo in other packages. Take a look at how we achieve that today.

Im for ditching rush, but dont want to spend too much time on that if it aint broken.

saskliutas:
I am interested in how this works? Should new apps based on this template be created using `degit iTwin/viewer/packages/templates/vite-template-web-viewer`?

I think `template.json` is a thing from CRA so I don't think it should be used anymore. From my experience with vite templates it should by buildable working thing that does not need any additional scripts and you could just copy paste. That's basically what `degit` is doing.

FYI, I have also built template based on vite https://github.com/saskliutas/vite-itwin-web-viewer.  Additionally I tried to cleanup and improve some React components used in template app. Maybe some of it could be included into official template.

ben-polinsky:
> We already re-use packages within the monorepo in other packages. Take a look at how we achieve that today.

The issue is that we're not setting `workspace:*` and rush detects this and throws an error, wanting us to use the local package. It's bonkers.

saskliutas:
> > We already re-use packages within the monorepo in other packages. Take a look at how we achieve that today.
> 
> The issue is that we're not setting `workspace:*` and rush detects this and throws an error, wanting us to use the local package. It's bonkers.

I guess it's because rush wants to keep same versions of dependencies between packages in monorepo. I would expect that this can be controlled through some setting.

Allowing to use different versions of the same dependency between packages in monorepo also has it's own problems and build failures especially if node_modules hoisting is not disabled.

aruniverse:
> > > We already re-use packages within the monorepo in other packages. Take a look at how we achieve that today.
> 
> > 
> 
> > The issue is that we're not setting `workspace:*` and rush detects this and throws an error, wanting us to use the local package. It's bonkers.
> 
> 
> 
> I guess it's because rush wants to keep same versions of dependencies between packages in monorepo. I would expect that this can be controlled through some setting.
> 
> 
> 
> Allowing to use different versions of the same dependency between packages in monorepo also has it's own problems and build failures especially if node_modules hoisting is not disabled.

I don't understand why were not using workspace:* in the template. When the template gets published to npm it'll fill in the exwcr version number.

saskliutas:
> > > > We already re-use packages within the monorepo in other packages. Take a look at how we achieve that today.
> > 
> > 
> > > 
> > 
> > 
> > > The issue is that we're not setting `workspace:*` and rush detects this and throws an error, wanting us to use the local package. It's bonkers.
> > 
> > 
> > I guess it's because rush wants to keep same versions of dependencies between packages in monorepo. I would expect that this can be controlled through some setting.
> > Allowing to use different versions of the same dependency between packages in monorepo also has it's own problems and build failures especially if node_modules hoisting is not disabled.
> 
> I don't understand why were not using workspace:* in the template. When the template gets published to npm it'll fill in the exact version number.

From my experience we do not publish template to npm. Vite does not have a bootstrapper like CRA that can bootstrap an application from template package. Their workflow is that templates are hosted on github for example and user use `degit` tool to copy/paste that onto their machine: https://vite.dev/guide/#community-templates

saskliutas:
> We do publish the CRA templates to npm, see https://www.npmjs.com/package/@itwin/cra-template-web-viewer

I know. That's because CRA had bootstrapper for creating applications from templates. Vite has `create-vite` tool but it only bootstraps applications from predefined templates. https://github.com/vitejs/vite/tree/main/packages/create-vite

I haven't tried but maybe `degit` supports copying subfolders from the repo. That way template could be kept where it is now. Only problem is that we need to specify viewer version using ^5.0.0 but I think we need to use local version. pnpm has this settings for achieving that. https://pnpm.io/settings#linkworkspacepackages. Not sure about rush.

pankhur94:
> I am interested in how this works? Should new apps based on this template be created using `degit iTwin/viewer/packages/templates/vite-template-web-viewer`?
> 
> I think `template.json` is a thing from CRA so I don't think it should be used anymore. From my experience with vite templates it should by buildable working thing that does not need any additional scripts and you could just copy paste. That's basically what `degit` is doing.
> 
> FYI, I have also built template based on vite https://github.com/saskliutas/vite-itwin-web-viewer. Additionally I tried to cleanup and improve some React components used in template app. Maybe some of it could be included into official template.

I've added `@itwin/web-viewer-react` to `decoupledLocalDependencies` in our rush configuration for our template. This allows it to be installed locally when running rush install, avoiding the workspace conflict error we previously encountered.

Additionally:

- The README has been updated to reflect usage of the new template.

- Components from @saskliutas’s repo have been included, as they were well-structured, with a few enhancements made on top.

packages/templates/vite-template-web-viewer/README.md
```diff
@@ -1,3 +1,15 @@
+# iTwin Viewer Create React App Template for Web
```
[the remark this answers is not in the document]
pankhur94: updated our Readme

rush.json
```diff
@@ -34,10 +34,11 @@
       "shouldPublish": true
     },
     {
-      "packageName": "@itwin/cra-template-web-viewer",
-      "projectFolder": "packages/templates/cra-template-web-viewer",
+      "packageName": "@itwin/vite-template-web-viewer",
+      "projectFolder": "packages/templates/vite-template-web-viewer",
       "versionPolicyName": "independentModule",
-      "shouldPublish": true
+      "shouldPublish": true,
```
aruniverse, line 40: if were going to use degit we dont need to publish anymore, set this to false

packages/templates/vite-template-web-viewer/package.json
```diff
@@ -0,0 +1,101 @@
...
+    "react",
+    "template",
+    "typescript",
+    "iModel",
+    "iTwinJs",
+    "iTwin",
+    "iTwin Viewer",
+    "vite"
+  ],
+  "description": "The iTwin Viewer template for creating web applications using vite.",
```
aruniverse, line 14:
```suggestion
  "description": "Based on the iTwin Viewer template for creating web applications using vite.",
```

packages/templates/vite-template-web-viewer/package.json
```diff
@@ -0,0 +1,101 @@
...
+  "repository": {
```
aruniverse, line 15: this wont make sense for a template

packages/templates/vite-template-web-viewer/package.json
```diff
@@ -0,0 +1,101 @@
...
+    "type": "git",
+    "url": "https://github.com/itwin/viewer.git",
+    "directory": "packages/modules/vite-template-web-viewer"
+  },
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/itwin/viewer/issues"
+  },
+    "dependencies": {
```
aruniverse, line 24:
```suggestion
  "dependencies": {
```

packages/templates/vite-template-web-viewer/package.json
```diff
@@ -0,0 +1,101 @@
...
+    "@itwin/presentation-shared": "^1.2.0",
+    "@itwin/unified-selection": "^1.3.0",
+    "@itwin/property-grid-react": "^1.16.0",
+    "@itwin/tree-widget-react": "^3.0.0",
+    "@itwin/web-viewer-react": "^5.0.0",
+    "@tanstack/react-router": "^1.124.0",
+    "@itwin/webgl-compatibility": "^5.0.0",
+    "@itwin/unified-selection-react": "^1.0.0",
+    "@remix-run/router": "^1.3.2",
+    "@types/react": "^18.2.18",
```
aruniverse, line 60: unlike cra which wouldnt allow for dev deps in its templates, we should be able to for degit + vite
also alphabetize the deps

packages/templates/vite-template-web-viewer/tsconfig.json
```diff
@@ -13,8 +13,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
aruniverse, line 16: lets update target to es2023

packages/templates/vite-template-web-viewer/src/components/Viewer.tsx
```diff
@@ -0,0 +1,123 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ECSchemaRpcInterface } from "@itwin/ecschema-rpcinterface-common";
```
aruniverse, line 6: i dont think we need to import this ourselves in the template
the viewer should already be registering this under the hood as of 5.0 now

common/changes/@itwin/vite-template-web-viewer/pankhur94-web-viewer-to-vite_2025-07-11-20-09.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
```
aruniverse, line 2: if we dont need to publish the pkg for vite then we dont need this change file

rush.json
pankhur94: set to false

packages/templates/vite-template-web-viewer/package.json
pankhur94: I have removed the fields which were only necessary when we publish the packages including keywords, description, repository, license, bugs. Also, shouldn't we update the versions of this package to something similar to web-viewer-test/desktop-viewer-test like "0.1.0" since we are not publishing this anymore?

packages/templates/vite-template-web-viewer/tsconfig.json
pankhur94: done

common/changes/@itwin/vite-template-web-viewer/pankhur94-web-viewer-to-vite_2025-07-11-20-09.json
pankhur94: wouldn't we need this to avoid error while running rush change in ci? I was hoping that changing `shouldPublish` to false would not require us to include this file when running the change command, but apparently it does.

packages/templates/vite-template-web-viewer/README.md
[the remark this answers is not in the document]
pankhur94: I agree but we should also let them know how to scaffold via degit. Shouldn't we let [this](https://github.com/iTwin/viewer/blob/pankhur94/web-viewer-to-vite/packages/templates/web/README.md#itwin-viewer-template-for-web-using-vite) section be there in our readme?

packages/templates/web/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "itwin-web-viewer-template",
+  "version": "5.0.0",
```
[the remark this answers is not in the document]
aruniverse: the version here doesnt matter

hl662:
Where are the icons? 
image

packages/templates/vite-template-web-viewer/README.md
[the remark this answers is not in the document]
aruniverse: By the time theyve scaffolded the app and this entire directory gets cloned on their end, theyd have already used degit. Youre right we need to have a section in our readme to show users how to use degit, but it should probably now be the one at the root, not the template readme. The template readme made more sense when it was published as a pkg on npm

hl662:
> Since the Vite config is copied over during scaffolding, I don’t think we should modify it just to accommodate this. Instead, we can continue testing the app locally by building and using vite preview

After scaffolding, will the developer see missing icons if they used pnpm/rushx start instead of preview? If they do, maybe hoist the package by customizing server.fs.allow [conditionally](https://vite.dev/config/#configuring-vite) when using vite serve

packages/templates/web/.env
```diff
@@ -0,0 +1,10 @@
+# ---- Authorization Client Settings ----
+IMJS_AUTH_CLIENT_CLIENT_ID = ""
+IMJS_AUTH_CLIENT_REDIRECT_URI = ""
+IMJS_AUTH_CLIENT_LOGOUT_URI = ""
+IMJS_AUTH_CLIENT_SCOPES ="itwin-platform"
+IMJS_AUTH_AUTHORITY="https://ims.bentley.com"
```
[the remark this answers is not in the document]
hl662: > but users will need to obtain their own key

Can that be documented? Perhaps a but users will need to obtain their own key to our docs on how to pass in their own key and start up google maps? 

P.S: Our documentation on Google Maps should be moved to a separate learnings doc, not just be in the changelog

saskliutas:
> > @saskliutas while the changes in the web template you made look fine, it does make exporting sandboxes and copying code from it to these templates a bit more difficult
> > unless we start updating all the sandboxes, which is is on our eventual todo list
> 
> Didn;t think about such use case. I will try to restructure a bit to make it more convenient to copy over code.

Tried to restructure my template a little. (Also included some changes from this PR). https://github.com/saskliutas/vite-itwin-web-viewer/tree/better-alignment-with-sandboxes It's still not straight copy paste from sandbox but `App.tsx` uses `Viewer` directly. Main change is how auth is handled. So it's up to you to decide if template should match sandboxes 1 to 1 or not.

aruniverse:
> > > @saskliutas while the changes in the web template you made look fine, it does make exporting sandboxes and copying code from it to these templates a bit more difficult
> > > unless we start updating all the sandboxes, which is is on our eventual todo list
> > 
> > 
> > Didn;t think about such use case. I will try to restructure a bit to make it more convenient to copy over code.
> 
> Tried to restructure my template a little. (Also included some changes from this PR). [saskliutas/vite-itwin-web-viewer@`better-alignment-with-sandboxes`](https://github.com/saskliutas/vite-itwin-web-viewer/tree/better-alignment-with-sandboxes?rgh-link-date=2025-07-15T15%3A28%3A20Z) It's still not straight copy paste from sandbox but `App.tsx` uses `Viewer` directly. Main change is how auth is handled. So it's up to you to decide if template should match sandboxes 1 to 1 or not.

This updated template looks good. There is still a divergence from the sandbox but I think we'll have to iterate on it a bit more. Sandbox will eventually need to start using public share keys so the auth would diverge from our templates a bit more.

@pankhur94 plz pull these changes in, and test it out. Also make sure the icons work.

packages/templates/web/.env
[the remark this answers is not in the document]
hl662: I opened a issue for [this](https://github.com/iTwin/viewer/issues/380), resolving this comment...

pankhur94:
> 
> This updated template looks good. There is still a divergence from the sandbox but I think we'll have to iterate on it a bit more. Sandbox will eventually need to start using public share keys so the auth would diverge from our templates a bit more.
> 
> @pankhur94 plz pull these changes in, and test it out. Also make sure the icons work.

I pulled in the changes and tested and things look good to me. Let me know if you we can merge this one.

PR #384, opened 2025-07-22
bump eslint-config-prettier - CVE-2025-54313 

ben-polinsky:
Though the malicious packages have been removed from NPM, out of caution we're bumping estlint-config-prettier above affected version.


ben-polinsky:
(and https://github.com/advisories/GHSA-fjxv-7rqg-78g4)

PR #392, opened 2025-08-08
Rename vite config to .mts to meet ESM, as vite's config resolver is ESM first

hl662:


hl662:
@saskliutas I don't know if you would have to update your template, we ran into this yesterday and renaming it fixed the issue.

Vite's config resolver couldn't find the file when it's as `.ts`

MichaelSwigerAtBentley (approved):

PR #406, opened 2025-10-03
use act from testing lib

aruniverse:


ben-polinsky (approved):

calebmshafer:
@aruniverse @ben-polinsky do we still want to add this to the tests?

PR #411, opened 2026-01-14
Pre-fill redirect uri for viewer templates

anmolshres98:
## Summary

Pre-fills the OAuth redirect URI in the web and desktop Viewer templates so scaffolded apps sign in out of the box without manual setup.

## Changes

- `templates/web/.env`: `IMJS_AUTH_CLIENT_REDIRECT_URI` → `http://localhost:3000/signin-callback`
- `templates/desktop/.env`: `ITWIN_VIEWER_REDIRECT_URI` → `http://localhost:3000/signin-callback`

> Register `http://localhost:3000/signin-callback` as an allowed redirect URI (or override for non-default ports).


ben-polinsky (approved):

calebmshafer:
@anmolshres98 @ben-polinsky what is the reason for pinning the caniuse-lite dependency?

Also, is the prefilling the redirect uri to ensure people don't need to copy/set it when doing client registration?

ben-polinsky:
> Also, is the prefilling the redirect uri to ensure people don't need to copy/set it when doing client registration?

This, I remember - yes.

anmolshres98:
> what is the reason for pinning the caniuse-lite dependency?

i believe it was because the audit was failing from a cve but I can take another pass soon and possibly remove the pin since it's been a while

anmolshres98:
@calebmshafer removed pin, updated description

ben-polinsky:
@anmolshres98 if you get a chance can you give this a once over?

anmolshres98:
> @anmolshres98 if you get a chance can you give this a once over?

this branch should be good, ci fails because of audit error existing in master. I will create separate PR with fix for those

calebmshafer (approved):

anmolshres98:
> > @anmolshres98 if you get a chance can you give this a once over?
> 
> this branch should be good, ci fails because of audit error existing in master. I will create separate PR with fix for those

https://github.com/iTwin/viewer/pull/430

